Issue #262, opened 2016-07-05
Force dependency installs


sg-:
mbed tools and the dependencies it relies on are changing very quickly adding support and fixes. Running `pip install requirements.txt` generally happens once and is the forgotten about, not every time the OS is pulled and updated.

mbed CLI should check and force the install (upgrade or downgrade) based on the version of OS and tools that are checked.

PR #280, opened 2016-07-17
Auto-install dependencies

screamerbg:
Behavior:
- mbed CLI will attempt to auto-install dependencies. If auto-install fails it will report error and provide instructions how to install #262
- Commands `compile`, `test`, `export` and `detect` now trigger dependencies auto-install when missing and also report error if unable to auto-install

Bump version to 0.9.0


BlackstoneEngineering:
wtf? this was awesome, why did this get closed? The autoinstalling of dependencies that the tools require is amazing, under this system when i ran compile or test or whatever the shit i needed would get installed. but under the current system it just breaks and doesnt do a damn thing.

PR #307, opened 2016-08-02
Editing

iriark01:
I've done quite a big edit, but most of what I changed is the order. I started with the sections people are likely to care about the very first time they use mbed CLI, and moved to the end the sections that will take a couple of days to reach (like updating libraries). I also put the sections in the order we'd probably like to encourage people to work at, so for example testing comes before publishing.

I’m assuming you can generate the ToC automatically. 

I had a few questions:
1 .Why do I have to install both Git and Mercurial? Why can’t I pick one?
2. For the status check before publishing, what do M and ?? mean, and are there other options?
3. Why do we tell people to manually edit the path to the compilers if we have `ARM_PATH`, `GCC_ARM_PATH`, `IAR_PATH` - defines the default path to ARM Compiler, GCC ARM and IAR Workbench toolchains. Default: none.?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Introduction
+## Introduction
 
 *mbed CLI* is the name of the ARM mbed command line tool, packaged as mbed-cli, which enables the full mbed workflow: repositories version control, maintaining dependencies, publishing code, updating from remotely hosted repositories (GitHub, GitLab and mbed.org), and invoking ARM mbed's own build system and export functions, among other operations.
 
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ But mbed CLI goes much further than the basic workflow. To support long-term dev
 
 ## Installation
 
-mbed CLImbed CLI is supported on Windows, Linux and Mac OS X. We're keen to learn about your experience with mbed CLI on other operating systems at the [mbed CLI development page](https://github.com/ARMmbed/mbed-cli).
+mbed CLI is supported on Windows, Linux and Mac OS X. We're keen to learn about your experience with mbed CLI on other operating systems at the [mbed CLI development page](https://github.com/ARMmbed/mbed-cli).
 
 ### Requirements
 
@@ -65,7 +65,7 @@ mbed CLImbed CLI is supported on Windows, Linux and Mac OS X. We're keen to lear
     <span class="tips">**Note:** The directories of Git and Mercurial executables (`git` and `hg`) need to be in your system's PATH.</span>
 
 * **Command-line compiler or IDE Toolchain** - mbed CLI invokes the [mbed OS 5](https://github.com/ARMmbed/mbed-os) tools for various features, like compiling, testing and exporting to industry standard toolchains. To compile your code, you will need either of these:
-    * Compilers: GCC ARM, ARMCC 5, IAR
+    * Compilers: GCC ARM, ARM Compiler 5, IAR
     * Toolchains: Keil uVision, DS-5, IAR Workbench
 
 ### Installing mbed CLI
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -36,88 +37,106 @@ This document covers the installation and usage of *mbed CLI*.
...
+1. Publish your application.
+
+But mbed CLI goes much further than the basic workflow. To support long-term development, mbed CLI offers nuanced source control, including selective updates of libraries and the code base, support for multiple toolchains, and manual configuration of the system.
+
+<span class="tips">**Tip:** mbed CLI help: To list all mbed CLI commands, use `mbed --help`. A detailed command-specific help is available by using `mbed <command> --help`.</span>
+
 ## Installation
 
-*mbed CLI* is supported on Windows, Linux and Mac OSX. We're keen to learn about your experience with *mbed CLI* on other operating systems at the [mbed CLI development page](https://github.com/ARMmbed/mbed-cli).
+mbed CLImbed CLI is supported on Windows, Linux and Mac OS X. We're keen to learn about your experience with mbed CLI on other operating systems at the [mbed CLI development page](https://github.com/ARMmbed/mbed-cli).
```
bridadan: `mbed CLImbed CLI` looks a little cramped and repetitive :)

README.md
```diff
@@ -424,34 +296,28 @@ Total Flash memory (text + data + misc): 66014 bytes
...
-* `--build <BUILD>` to select the build directory. Default: `.build/` inside your program
-* `--library` to compile the code as a [static .a/.ar library](#compiling-static-libraries)
-* `--config` to inspect the run-time compile configuration (see below)
-* `-S` or `--supported` shows supported matrix of targets and toolchains
-* `-c ` (optional): will build from scratch; a clean build or rebuild.
-* `-j <jobs>` (optional): control the compile threads on your machine. Default value is 0, which infer the number of threads from the number of cores on your machine. You can use `-j 1` to trigger sequential compile of source code..
-* `-v` or `--verbose` for verbose diagnostic output
-* `-vv` or `--very_verbose` for very verbose diagnostic output
+* `-m <MCU>` to select a target.
+* `-t <TOOLCHAIN>` to select a toolchain (of those defined in `mbed_settings.py`, see above). The value can be either `ARM` (armcc compiler), `GCC_ARM` (GNU ARM Embedded), or `IAR` (IAR Embedded Workbench for ARM).
```
bridadan: Should we be calling the "ARM" compiler "armcc" or "ARM Compiler 5"?

screamerbg:
@iriark01 Thanks for this. Looks awesome!

Could you please fix the TOC (unfortunately it's not automatically generated) and address the minor issues @bridadan pointed?

iriark01:
Someone here does know how to auto-generate the ToC. Was it @meriac ?

iriark01:
Also, did you see the questions in my PR? @screamerbg

screamerbg:
@iriark01 
1 .Why do I have to install both Git and Mercurial? Why can’t I pick one?
 * All mbed 2.0 programs and libraries are hosted on mbed.org and based on Mercurial. mbed OS and related libraries are hosted on Github and based on Git. It's unrealistic to consider that only Git will be used.

2. For the status check before publishing, what do M and ?? mean, and are there other options?
 * What line of the documentation is this?

3. Why do we tell people to manually edit the path to the compilers if we have `ARM_PATH`, `GCC_ARM_PATH`, `IAR_PATH` - defines the default path to ARM Compiler, GCC ARM and IAR Workbench toolchains. Default: none.?
 * Correct. We should rather point users at setting up the path for their toolchains via: `mbed config -G ARM_PATH=...`, `mbed config -G GCC_ARM_PATH=....` and `mbed config -G IAR_PATH=...`

screamerbg:
Whops, sorry for closing it. Wrong click

meriac:
@iriark01 on github I used in the past Jekyll to auto-generate TOC's on github pages - but the toc will be only visible in Jekyll-version of the page. Ithi might be still helpful though for pulling in such pages automatically - I would just use a plain template without footers and headers.

PR #316, opened 2016-08-04
Three changes

iriark01:
1. Added a simplified ToC (only level two headers); this is partially cause I don't have time to build the level two into it, and partially because on docs.mbed it's cluttering the doc without being helpful, as there already is an auto-generated ToC
2. Changed mbed-blinky to mbed-os-example-blinky
3. Added a tip about the auto-copmlete of https://github.com/ARMmbed/ (thanks to Bogdan)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,17 +6,37 @@ This document covers the installation and usage of mbed CLI.
 
 ## Table of Contents
 
-1. [Using mbed CLI](#using-mbed-cli)
-1. [Installation](#installation)
-1. [Before you begin](#before-you-begin-understanding-the-working-context-and-program-root)
+1. [Requirements](#requirements)
+1. [Installing and uninstalling](#installing-mbed-cli)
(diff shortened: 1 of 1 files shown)


screamerbg:
@iriark01 thanks. Regarding item 1 - docs.mbed is not the only entry point to the mbed CLI guide and I'd prefer to keep the toc here. Also the toc could be hidden on docs.mbed if certain css class is used. Perhaps @JoeTheGuitarist could advise.

iriark01:
As I said, it's also about time. I don't have time to build the second level into it manually.

screamerbg:
@iriark01 can you update your PR so the toc is unchanged then? I'll update it later today

iriark01:
Why? This version is more correct than the one currently there.

screamerbg:
The current version has 4 outdated links. Your PR removes 18 working links.

screamerbg:
Thanks

PR #345, opened 2016-09-13
Adustm issue342: add a new detail in the readme file

adustm:

Issue #342, opened 2016-09-12
mbed test - enhancement request - Could we limit the scope to a test group ?


adustm:
Hello, I've been reading https://github.com/ARMmbed/mbed-cli#limiting-the-test-scope

We would be happy to have the possibility to launch a tests family without adding the complete test names of each test_case.

Do you think it is feasible ?

Kind regards
Armelle
cc @svastm @bcostm

bridadan:
@adustm There is already the capability to launch a family of tests that have the same prefix by using a `*` character. You can run `mbed test -n tests-mbedmicro-rtos*` for instance to run all RTOS tests.

There has been ongoing work on Greentea to add support for "Test plans" that should provide more flexibility for executing a list of tests. The PR is open here: https://github.com/ARMmbed/greentea/pull/175

Would either of these features work for you?

adustm:
Hello Brian, 
Thank your for the trick, that's great and work fine.

I have created a PR to add this detail in the readme file, if you don't mind:
https://github.com/ARMmbed/mbed-cli/pull/345
You can use it directly or rephrase it the way you like, but could you please add this in the readme before I close this issue ?
Thanks in advance
Armelle

bridadan:
(Copying this here as well for clarity)

I made an alternative for this PR to clarify a few points, would mind taking a look here: #346

adustm:
LGTM

PR #346, opened 2016-09-13
Adding details on wildcard character for limiting tests

bridadan:
This is an alternative PR for @adustm's PR https://github.com/ARMmbed/mbed-cli/pull/345 in response to issue #342.

It adds a bit more detail and warns of some shells expanding the `*` character into local file names.


screamerbg:
LGTM

adustm:
Thank you Brian, I'll close the issue

PR #349, opened 2016-09-16
Changing CI to use available test in latest mbed-os

bridadan:
The test called `tests-integration-threaded_blinky` was removed from mbed OS recently. I've replaced the line in CI that compiled it with an available test.


screamerbg:
Great, thanks!

bridadan:
@screamerbg According to the CI it still didn't find the test when trying to build it:

```
$ cd .tests/new-test && mbed test --compile -n tests-integration-basic -j 0
[Warning] Test with name 'tests-integration-basic' was not found in the available tests
...
```

I'm pretty sure the test name needs to be prefixed with `mbed-os-`, I will add this in another PR

PR #365, opened 2016-10-06
Correcting parameters '-o' and '--app-config'

bridadan:
This PR makes two changes with regards to command line arguments:

1) The `--app-config` parameter was wrongfully being passed to `mbedgt`. It is only a valid parameter for the `test.py` script. It has been removed.

2) The `-o` parameter has been replaced with the more general and flexible `--profile` parameter.

## TODO
- [x] Review by @screamerbg 
- [ ] Review by @theotherjimmy


screamerbg:
LGTM. Thanks.

PR #366, opened 2016-10-07
Use 'urllib2' for retrieving mbed library

betzw:
For some reason when using `urllib` in my environment (which includes a Zscaler network proxy) `urllib.urlretrieve()` does not work ending up with an error message _"301 Moved Permanently"_.
Python library `urllib2` doesn't have this issue but does not provide method `urlretrieve()`, so a little workaround is required.


screamerbg:
Thanks! LGTM

PR #373, opened 2016-10-15
Corrected type in pushing upstream example

sillevl:
Text speaks of 'ot-client', but should be 'iot-client' as in the code example above.


screamerbg:
@sillevl Thanks

PR #381, opened 2016-10-24
Use developer.mbed.org to avoid redirect

stephenpaulger:
Using mbed.org means a HTTP redirect must be followed, so using developer.mbed.org directly should make requests marginally faster.


theotherjimmy:
@stephenpaulger Should this be changed to `os.mbed.com`?

stephenpaulger:
Ideally yes.

screamerbg:
@stephenpaulger Are you happy if we close this?

stephenpaulger:
It's somewhat out of date now.

PR #383, opened 2016-10-28
Changing ToC indents

iriark01:
The docs.mbed can't see your indents, so the ToC is 27 flat items. It should be able to see these indents and given properly nested bullets.


screamerbg:
Thanks

PR #405, opened 2016-12-12
Update README.md

AnotherButler:
Made Irit-approved grammar changes [mostly changing passive voice to active voice and fixing hyphens]

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ## Introduction
 
-*mbed CLI* is the name of the ARM mbed command-line tool, packaged as mbed-cli, which enables the full mbed workflow: repositories version control, maintaining dependencies, publishing code, updating from remotely hosted repositories (GitHub, GitLab and mbed.org), and invoking ARM mbed's own build system and export functions, among other operations.
+*mbed CLI* is the name of the ARM mbed command-line tool, packaged as mbed-cli, which enables the full mbed workflow: repositories version control, maintaining dependencies, publishing code, updating from remotely hosted repositories (GitHub, GitLab and mbed.org) and invoking ARM mbed's own build system and export functions, among other operations.
 
 This document covers the installation and usage of mbed CLI.
 
@@ -54,7 +54,7 @@ Windows, Linux and Mac OS X support mbed CLI. We're keen to learn about your exp
 
 ### Requirements
 
-* **Python** - mbed CLI is a Python script, so you'll need Python to use it. We tested mbed CLI with [version 2.7.11 of Python](https://www.python.org/downloads/release/python-2711/). It is not compatible with Python 3.
+* **Python** - mbed CLI is a Python script, so you'll need Python to use it. We test mbed CLI with [version 2.7.11 of Python](https://www.python.org/downloads/release/python-2711/). It is not compatible with Python 3.
 
 * **Git and Mercurial** - mbed CLI supports both Git and Mercurial repositories, so you'll need to install both:
     * [Git](https://git-scm.com/) - version 1.9.5 or later.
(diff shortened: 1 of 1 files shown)


AnotherButler:
Under "Creating a new program for mbed OS 2," both links in the following sentence go to the same place: 

"mbed CLI is also compatible with mbed OS 2 programs based on the mbed library, and will automatically import the latest mbed library release if you use the --mbedlib option:"

Should one of these links go somewhere else?

README.md
```diff
@@ -44,27 +44,27 @@ The basic workflow for mbed CLI is to:
...
 
 ## Installation
 
-mbed CLI is supported on Windows, Linux and Mac OS X. We're keen to learn about your experience with mbed CLI on other operating systems at the [mbed CLI development page](https://github.com/ARMmbed/mbed-cli).
+Windows, Linux and Mac OS X support mbed CLI. We're keen to learn about your experience with mbed CLI on other operating systems at the [mbed CLI development page](https://github.com/ARMmbed/mbed-cli).
 
 ### Requirements
 
-* **Python** - mbed CLI is a Python script, so you'll need Python in order to use it. mbed CLI was tested with [version 2.7.11 of Python](https://www.python.org/downloads/release/python-2711/) and is not compatible with Python 3.
+* **Python** - mbed CLI is a Python script, so you'll need Python to use it. We tested mbed CLI with [version 2.7.11 of Python](https://www.python.org/downloads/release/python-2711/). It is not compatible with Python 3.
```
bridadan: We `tested` and we also currently `test` with Python 2.7.x, past and present tense. I don't mind which tense we use, just thought I'd point out we're also continuing to test it :)

README.md
```diff
@@ -44,27 +44,27 @@ The basic workflow for mbed CLI is to:
...
-    * [Mercurial](https://www.mercurial-scm.org/) - version 2.2.2 or later
+    * [Git](https://git-scm.com/) - version 1.9.5 or later.
+    * [Mercurial](https://www.mercurial-scm.org/) - version 2.2.2 or later.
 
     <span class="tips">**Note:** The directories of Git and Mercurial executables (`git` and `hg`) need to be in your system's PATH.</span>
 
-* **Command-line compiler or IDE Toolchain** - mbed CLI invokes the [mbed OS 5](https://github.com/ARMmbed/mbed-os) tools for various features, like compiling, testing and exporting to industry standard toolchains. To compile your code, you will need either of these:
-    * Compilers: GCC ARM, ARM Compiler 5, IAR
-    * Toolchains: Keil uVision, DS-5, IAR Workbench
+* **Command-line compiler or IDE toolchain** - mbed CLI invokes the [mbed OS 5](https://github.com/ARMmbed/mbed-os) tools for various features, such as compiling, testing and exporting to industry standard toolchains. To compile your code, you will need a compiler and a toolchain:
```
bridadan: You don't need both the compiler and a toolchain, you need either a compiler *or* a toolchain.

In all honesty, we use the terms `compiler` and `toolchain` slightly interchangeably. I would personally recommend changing the word `toolchain` to `IDE`. If this requires changes through the rest of the document to keep this consistent, then that can come in a separate PR.

README.md
```diff
@@ -763,27 +763,27 @@ Run `mbed update <tag_name|revision> --clean`
 
 __When you have unpublished local libraries__
 
-There are three additional options that define how unpublished local libraries are handled:
+There are three additional options that show how unpublished local libraries are handled:
```
bridadan: Might just be me but when I see `show` coupled with an option description, I assume it's going to print something to screen. When really these options modify the behavior of the tool. Perhaps we can say `modify` instead of `define` or `show`?

bridadan:
Changes look good to me! Care to take a look @screamerbg?

screamerbg:
Thanks!

PR #423, opened 2017-01-17
Remove unnecessary creation of BUILD dir for certain commands.

### LINKED ISSUE #413: BUILD folder always created
When using --build option in "mbed test" command (maybe happen when using other commands too), BUILD folder is also created.
It contains only .mbedignore file.
Happen because of this line: 
program.ignore_build_dir() in def test_ (from mbed_test.py)

bridadan:
mbed CLI was creating the BUILD directory even in cases where it wasn't needed. This PR defers the creation of the BUILD dir until before compile steps only.

This should fix #413.

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -2144,7 +2144,6 @@ def compile_(toolchain=None, target=None, profile=False, compile_library=False, 
     # Find the root of the program
     program = Program(os.getcwd(), True)
     program.check_requirements(True)
-    program.ignore_build_dir()
     # Remember the original path. this is needed for compiling only the libraries and tests for the current folder.
     orig_path = os.getcwd()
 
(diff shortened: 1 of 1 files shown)


mbed/mbed.py
```diff
@@ -2178,39 +2177,44 @@ def compile_(toolchain=None, target=None, profile=False, compile_library=False,
...
+            # Compile as application (root is default)
+            if not build:
+                build = os.path.join(os.path.relpath(program.path, orig_path), program.build_dir, target, tchain)
+
+            popen(['python', '-u', os.path.join(tools_dir, 'make.py')]
+                  + list(chain.from_iterable(izip(repeat('-D'), macros)))
+                  + ['-t', tchain, '-m', target]
+                  + list(chain.from_iterable(izip(repeat('--profile'), profile or [])))
+                  + list(chain.from_iterable(izip(repeat('--source'), source)))
+                  + ['--build', build]
```
screamerbg: Can you align `build` with `build_path` naming below for consistency?

bridadan:
Thanks for the feedback @screamerbg, could you take a look again?

screamerbg:
Thanks :)

PR #438, opened 2017-02-22
Dual Python 2 and 3 support (not ready to merge)

danielward:
Currently the client only supports Python 2. The intention of this pull request is to introduce support for Python 3, whilst continuing support for Python 2 to minimise compatibility issues.


danielward:
Removing this pending solutions in dependencies.

Issue #458, opened 2017-03-22
'mbed import' doesn't have a consistent behaviour across multiple executions


MarceloSalazar:
This is considered a bug

mbed CLI: 1.0.0
Windows 10
Python: 2.7.13

Summary:
1. Import blinky (I get mbed OS 5.4.1)
2. Update blinky project to mbed OS 5.3.6
3. Move to other directory and import blinky2 (I get mbed mbed OS 5.3.6, whether I should get 5.4.1 - latest!).

Logs:
```
C:\t>mbed import mbed-os-example-blinky
[mbed] Importing program "mbed-os-example-blinky" from "https://github.com/ARMmbed/mbed-os-example-blinky" at latest revision in the current branch
[mbed] Adding library "mbed-os" from "https://github.com/ARMmbed/mbed-os" at rev #f4864dc6429e

C:\t>cd mbed-os-example-blinky

C:\t\mbed-os-example-blinky>mbed ls -a
mbed-os-example-blinky (https://github.com/ARMmbed/mbed-os-example-blinky#9dbaf52e7d2a)
`- mbed-os (https://github.com/ARMmbed/mbed-os#f4864dc6429e)

C:\t\mbed-os-example-blinky>mbed update mbed-os-5.3.6     # manually switch to old version
[mbed] Updating program "mbed-os-example-blinky" to branch mbed-os-5.3.6
[mbed] Updating library "mbed-os" to rev #5faf4b26c595

C:\t\mbed-os-example-blinky>mbed ls -a
mbed-os-example-blinky (https://github.com/ARMmbed/mbed-os-example-blinky#4d23fd49c818)
`- mbed-os (https://github.com/ARMmbed/mbed-os#5faf4b26c595)

C:\t\mbed-os-example-blinky>cd ..

C:\t>mbed import mbed-os-example-blinky blinky2           # import new project
[mbed] Importing program "blinky2" from "mbed-os-example-blinky" at latest revision in the current branch
Note: checking out '4d23fd49c8182620950e7bfe6edd6a9b58329513'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

[mbed] Adding library "mbed-os" from "https://github.com/ARMmbed/mbed-os" at rev #5faf4b26c595

[...]

PR #461, opened 2017-03-23
Show a warning when importing from a local folder

screamerbg:
This addresses #458


theotherjimmy:
👍 for the small change set.

PR #466, opened 2017-03-30
New Feature: Flash Target Board

screamerbg:
Added feature support for -f/--flash to `mbed compile` to flash (and reset) connected target board after successful compile.

The implementation uses greentea/htrun framework and therefore relies on the same routines as testing.

To use:
```
$ mbed compile -t <toolchain> -m <mcu|DETECT> -f|--flash
```

Example output:
```
Total Static RAM memory (data + bss): 22732 bytes
Total RAM memory (data + bss + heap + stack): 22732 bytes
Total Flash memory (text + data + misc): 30266 bytes

Image: .\BUILD\NUCLEO_F429ZI\ARM\mbed-os-example-blinky.bin
[mbed] Detected "NUCLEO_F429ZI" connected to "E:" and using com port "COM6"
        1 file(s) copied.
```

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1400,16 +1400,9 @@ class Program(object):
         target = target if target else target_cfg
 
         if target and (target.lower() == 'detect' or target.lower() == 'auto'):
-            targets = self.get_detected_targets()
-            if targets == False:
-                error("The target detection requires that the 'mbed-ls' python module is installed.\nYou can install mbed-ls by running 'pip install mbed-ls'.")
-            elif len(targets) > 1:
-                error("Multiple targets were detected.\nOnly 1 target board should be connected to your system when you use the '-m auto' switch.")
-            elif len(targets) == 0:
-                error("No targets were detected.\nPlease make sure a target board is connected to this system.")
-            else:
-                action("Detected \"%s\" connected to \"%s\" and using com port \"%s\"" % (targets[0]['name'], targets[0]['mount'], targets[0]['serial']))
-                target = targets[0]['name']
+            detected = self.detect_target()
+            if detected:
+                target = detected['name']
 
         if target is None:
             error("Please specify target using the -m switch or set default target using command 'mbed target'", 1)
@@ -1446,6 +1439,22 @@ class Program(object):
                     f.write('*\n')
             except IOError:
                 error("Unable to write build ignore file in \"%s\"" % os.path.join(build_path, '.mbedignore'), 1)
+
+    def detect_target(self, info=None):
+        targets = self.get_detected_targets()
+        if targets == False:
+            error("The target detection requires that the 'mbed-ls' python module is installed.\nYou can install mbed-ls by running 'pip install mbed-ls'.")
+        elif len(targets) > 1:
+            error("Multiple targets were detected.\nOnly 1 target board should be connected to your system.")
+        elif len(targets) == 0:
+            error("No targets were detected.\nPlease make sure a target board is connected to this system.")
+        else:
+            action("Detected \"%s\" connected to \"%s\" and using com port \"%s\"" % (targets[0]['name'], targets[0]['mount'], targets[0]['serial']))
+            info = {'msd': targets[0]['mount'], 'port': targets[0]['serial'], 'name': targets[0]['name']}
+
+        if info is None:
+            error("The detected target doesn't support Mass Storage Device capability (MSD)", 1)
+        return info
 
     def get_detected_targets(self):
         targets = []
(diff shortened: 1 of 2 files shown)


mbed/mbed.py
```diff
@@ -1447,6 +1440,22 @@ def ignore_build_dir(self):
...
                 error("Unable to write build ignore file in \"%s\"" % os.path.join(build_path, '.mbedignore'), 1)
 
+    def detect_target(self, info=None):
+        targets = self.get_detected_targets()
+        if targets == False:
+            error("The target detection requires that the 'mbed-ls' python module is installed.\nYou can install mbed-ls by running 'pip install mbed-ls'.")
+        elif len(targets) > 1:
+            error("Multiple targets were detected.\nOnly 1 target board should be connected to your system.")
+        elif len(targets) == 0:
+            error("No targets were detected.\nPlease make sure a target board is connected to this system.")
```
theotherjimmy: On linux, it's not enough to connect the target board to the system. You must also mount the fake file system of the board somewhere. Should we mention this in the error message?

mbed/mbed.py
```diff
@@ -2221,6 +2231,23 @@ def compile_(toolchain=None, target=None, profile=False, compile_library=False,
                   + (['-v'] if verbose else [])
                   + args,
                   env=env)
+            
+            if flash:
```
theotherjimmy: Before trying to flash, should you also check that the return code of the compile job indicates success?

mbed/mbed.py
```diff
@@ -2221,6 +2231,23 @@ def compile_(toolchain=None, target=None, profile=False, compile_library=False,
...
+                fw_name = artifact_name if artifact_name else program.name
+                fw_fbase = os.path.join(build_path, fw_name)
+                fw_file = fw_fbase + ('.hex' if os.path.exists(fw_fbase+'.hex') else '.bin')
```
theotherjimmy: You only check that the `.hex` file exists. I think you should check for `.bin` as well.

screamerbg:
@theotherjimmy Thanks for your comments. All of them make sense (and now addressed) 👍

mbed/mbed.py
```diff
@@ -2221,6 +2231,25 @@ def compile_(toolchain=None, target=None, profile=False, compile_library=False,
...
+                if not os.path.exists(fw_file):
+                    error("Build program file (firmware) not found \"%s\"" % fw_file, 1)
+                detected = program.detect_target()
+
+                try:
+                    from mbed_host_tests.host_tests_toolbox import flash_dev, reset_dev
+                except (IOError, ImportError, OSError):
+                    error("The '-f/--flash' option requires that the 'mbed-greentea' python module is installed.\nYou can install mbed-ls by running 'pip install mbed-greentea'.", 1)
```
bridadan: This is really part of the `mbed-host-tests` module, not `mbed-greentea`. `mbed-greentea` will correctly bring in this dependency though. So it's ok to leave this as is I think, but just thought I'd mention it.

mbed/mbed.py
```diff
@@ -2221,6 +2231,25 @@ def compile_(toolchain=None, target=None, profile=False, compile_library=False,
...
+
+                if not flash_dev(detected['msd'], fw_file, program_cycle_s=2):
+                    error("Unable to flash the target board connected to your system.", 1)
```
bridadan: You might also prompt the user to check for a `FAIL.TXT` on the board for more information. Not necessary though.

mbed/mbed.py
```diff
@@ -1447,6 +1440,22 @@ def ignore_build_dir(self):
             except IOError:
                 error("Unable to write build ignore file in \"%s\"" % os.path.join(build_path, '.mbedignore'), 1)
 
+    def detect_target(self, info=None):
```
bridadan: There is a valid case where you have multiple platforms plugged in, but all of them are different. This could be supported by supplying an optional target name to this function. Shouldn't be too hard to implement:

```python
def detect_target(self, target_name=None, info=None):
    targets = self.get_detected_targets()

    if target_name:
        targets = [t for t in targets if t['name'].upper() == target_name.upper()]

    ...
```

(^Note that actual code is totally untested, but the idea should be sound)

But this can always be added in another PR. @screamerbg totally up to you on the priority of this.

screamerbg: What's the use case for detecting multiple targets? Also what if I have multiple targets of the same kind mixed with different ones?

I guess I'm missing the point.

mbed/mbed.py
screamerbg: Why not have that check that in htrun instead, and have the top function handle it?

mbed/mbed.py
screamerbg: The developer never installs mbed-host-tests manually, neither are familiar with this module. Generally we would like to point developer at installing mbed-greentea as foundation of mbed OS testing, not specifically mbed-host-tests.

mbed/mbed.py
bridadan: Sure, also a possibility. Like I said, not necessary and should't block this PR.

mbed/mbed.py
bridadan: Makes sense to me!

mbed/mbed.py
bridadan: Say I have a LPC1768, a K64F, and an NRF51_DK plugged into my machine. I could run `mbed compile -m NRF51_DK ... --flash` and mbed CLI could correctly figure out which board to flash. Can be useful when developing for multiple platforms, but maybe not important as of right now.

> Also what if I have multiple targets of the same kind mixed with different ones?

In this case it would have to error like you do above since there are multiple boards (for example, 2 LPC1768s plugged in. You don't know which one to flash just by using the target name).

I'm afraid I'm hindering this PR more than I'm helping it with this comment. I'm happy to submit the change I mentioned above as a PR after this is merged for more discussion.

PR #470, opened 2017-03-30
Do not assume that python means python2

bremoran:
mbed-cli cannot assume that the system has python2 installed as python. This causes breakage for developers who have installed python3 prior to python2.

cc @screamerbg

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -916,7 +916,7 @@ class Repo(object):
 
         repo.path = os.path.abspath(path)
         repo.name = os.path.basename(repo.path)
-        
+
         cache_cfg = Global().get_cfg('CACHE', '')
         if cache_repositories and cache_cfg and cache_cfg != 'none' and cache_cfg != 'off' and cache_cfg != 'disabled':
             loc = cache_cfg if (cache_cfg and cache_cfg != 'on' and cache_cfg != 'enabled') else None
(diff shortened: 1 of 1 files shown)


mbed/mbed.py
```diff
@@ -916,7 +916,7 @@ def fromrepo(cls, path=None):
 
         repo.path = os.path.abspath(path)
         repo.name = os.path.basename(repo.path)
-        
+
```
theotherjimmy: :+1: for removing these spaces.

theotherjimmy:
Or people on archlinux.

theotherjimmy:
Does this work on windows?

bridadan:
Just typed in `python2` into my windows command prompt and it didn't bring anything up :/ I'm on 2.7.11

theotherjimmy:
Welp. That's that then. We cannot accept this patch as is.

bremoran:
I didn't realize that python installers on windows don't populate python2 and python3.

theotherjimmy:
@screamerbg Could you take a look at this patch? I think this is a step in the right direction.

screamerbg:
@bremoran This is great! Thanks for the PR! Could you please confirm that it works in venv as well?

@theotherjimmy Yes, I agree!

bridadan:
Before this gets released, it may be a good idea to let all the CI teams try out a release branch before putting this out in the wild. Specifically @tommikas, @mazimkhan, and @studavekar probably care about this :smile:

theotherjimmy:
Are we waiting on approval from @tommikas @mazimkhan and @studavekar then?

mazimkhan:
I would leave it to be tested in Oulu and Austin.

screamerbg:
@bremoran Thanks for this! Now going in mbed CLI v1.1

bridadan:
@screamerbg Do we have a timeline for the 1.1 release? I think I'd like to get this into the Austin CI at some point beforehand just to do a bit of soak testing.

PR #478, opened 2017-04-21
Fix virtualenv --relocatable issue

korjaa:
When setuptools creates console scripts during the install process to the host system and no shebang is defined in the setup.py, setuptools seems to hard code the current Python path as the shebang of the newly created script. This breaks the relocation feature in virtualenv. The issue is that when the virtualenv is moved to a new location the old absolute path in the console script shebang stays. By defining shebang in the setup.py, setuptools copies it to the console scripts it creates for the given module.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 # Copyright (c) 2016 ARM Limited, All Rights Reserved
 # SPDX-License-Identifier: Apache-2.0
 


korjaa:
Seems to work as expected:
```
mkdir 1
cd 1
virtualenv venv
source venv/bin/activate
pip install mbed-cli
deactivate
virtualenv --relocatable venv

cd ..
mv 1 2
cd 2
//fix activate script python path
source venv/bin/activate
```

korjaa:
Still requires the virtualenv path fix to `venv/bin/activate`

    VIRTUAL_ENV="$(readlink -f $(dirname $(readlink -f $BASH_SOURCE))/..)"

But that's not related to mbed-cli.

PR #483, opened 2017-04-25
README.md - cache none

JanneKiiskila:
Add example of none to turn cache off.


screamerbg:
Thanks for the contribution

PR #495, opened 2017-05-30
Update gcc version

theotherjimmy:
The docs should reflect our current standards :D


screamerbg:
LGTM

jeromecoutant:
Hi
Maybe you should also update:
https://docs.mbed.com/docs/mbed-os-handbook/en/latest/dev_tools/options/#compiler-versions

theotherjimmy:
@jeromecoutant We have made that update to the 5.5 docs (here: https://github.com/ARMmbed/Handbook/blob/5.5/docs/dev_tools/options.md) Which will go live when mbed OS 5.5 is released.

PR #512, opened 2017-06-15
Add -l/--latest-deps option to update

### LINKED ISSUE #511: Enhancement request: mbed upgrade or mbed update -f
See conversation over in #505, it is a complete pain trying to develop a set of libraries as there is no way to tell mbed to just update everything to latest.  I recognise that there are dangers in doing this but, when actively developing a suite of libraries, a developer is perfectly aware of any changes he/she has made and just wants everything to stay in sync with what's archived; the only way to do that is to jump through all the directories and update each one individually, which is just horrible, almost unworkable, especially when one has to do it in the right order to avoid elements higher in the try *downgrading* the ones below again.

Hence, I request a new feature, which could either be `mbed upgrade` or a `-f` (==force) switch to `mbed update`.  The effect of this would be to update all the libraries to their latest versions, re-writing the `.lib` files as it goes.  A warning could be provided, something like "All libraries will be update to their latest versions, which may introduce incompatibilities; are you sure you want to continue? (y/n)"

theotherjimmy:
Resolves #511

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1943,12 +1943,13 @@ def publish(all_refs=None, msg=None, top=True):
     dict(name=['-I', '--ignore'], action='store_true', help='Ignore errors related to unpublished libraries, unpublished or uncommitted changes, and attempt to update from associated remote repository URLs.'),
     dict(name='--depth', nargs='?', help='Number of revisions to fetch from the remote repository. Default: all revisions.'),
     dict(name='--protocol', nargs='?', help='Transport protocol for the source control management. Supported: https, http, ssh, git. Default: inferred from URL.'),
+    dict(name=['-l', '--latest-deps'], action='store_true', help='Update all dependencies to the latest revision of their current branch. WARNING: Ignores lib files'),
     help='Update to branch, tag, revision or latest',
     description=(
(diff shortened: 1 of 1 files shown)


screamerbg:
@theotherjimmy Like the size the patch (really awesome!!), but not sure about the switch name. 
`-r/--recursive` leaves the impression that `update` doesn't work recursively. But `update` with no parameter would only update the current repository item and align sub-repositories to match the correct hash by recursively update them as well.

bridadan:
@screamerbg Fair point. What if it was named `--recursively-update`? It's getting a bit lengthy now, but that may do a better job of describing the intended action.

theotherjimmy:
@screamerbg Do you have an alternative suggestion? I think `-f/--force` was suggested in #511, but that does not really explain itself well.

theotherjimmy:
`-u/--update-deps`?
`-i/--include-deps`?
`-i/--ignore-lib-files`?
`-m/--most-recent-deps`?

theotherjimmy:
I kinda like `-l/--latest-deps` so I rewrote history to pretend that's what I was suggesting all along. :smile: What do you think of this version @screamerbg

screamerbg:
And we have a winner gentleman :) LGTM

Issue #504, opened 2017-06-13
mbed export is missing parameter --app-config


AriParkkila:
For example _mbed-os-example-mesh-minimal_ need to be compiled with --app-config parameter. That parameter seems to be missing from mbed export command.

bridadan:
Thanks for reporting this @AriParkkila.

@theotherjimmy Do the tools support overriding the default filename `mbed_app.json` when exporting projects via the `--app-config` parameter?

theotherjimmy:
Let me check real quick.

theotherjimmy:
Not right now, but it would be a simple patch.

theotherjimmy:
```diff
diff --git i/tools/export/__init__.py w/tools/export/__init__.py
index e586c22b8..64dadf68f 100644
--- i/tools/export/__init__.py
+++ w/tools/export/__init__.py
@@ -255,7 +255,7 @@ def export_project(src_paths, export_path, target, ide, libraries_paths=None,
                    linker_script=None, notify=None, verbose=False, name=None,
                    inc_dirs=None, jobs=1, silent=False, extra_verbose=False,
                    config=None, macros=None, zip_proj=None, inc_repos=False,
-                   build_profile=None):
+                   build_profile=None, app_config=None):
     """Generates a project file and creates a zip archive if specified
 
     Positional Arguments:
@@ -307,7 +307,8 @@ def export_project(src_paths, export_path, target, ide, libraries_paths=None,
     toolchain = prepare_toolchain(
         paths, "", target, toolchain_name, macros=macros, jobs=jobs,
         notify=notify, silent=silent, verbose=verbose,
-        extra_verbose=extra_verbose, config=config, build_profile=build_profile)
+        extra_verbose=extra_verbose, config=config, build_profile=build_profile,
+        app_config=app_config)
     # The first path will give the name to the library
     if name is None:
         name = basename(normpath(abspath(src_paths[0])))
diff --git i/tools/project.py w/tools/project.py
index dce04ab64..d7185a78e 100644
--- i/tools/project.py
+++ w/tools/project.py
@@ -66,7 +66,8 @@ def setup_project(ide, target, program=None, source_dir=None, build=None, export
 
 
 def export(target, ide, build=None, src=None, macros=None, project_id=None,
-           zip_proj=False, build_profile=None, export_path=None, silent=False):
+           zip_proj=False, build_profile=None, export_path=None, silent=False,
+           app_config=None):
     """Do an export of a project.
 
[...]

PR #514, opened 2017-06-20
Adding --app-config option to 'mbed export'

bridadan:
This is the final piece to #504. It requires the yet to be release mbed-os 5.5.1 (capability added here: https://github.com/ARMmbed/mbed-os/pull/4540).

FYI @theotherjimmy


screamerbg:
LGTM

PR #515, opened 2017-06-21
Ignore root makefiles only

### LINKED ISSUE #490: "Makefile" added into GIT repo excludes
It seems that mbed-cli adds `Makefile` (and others) to `.git/info/exclude`. What's the reason to force repository wide ignore of those? 

I presume it is because of project export, if so I think these ignores could be more specific so that user is allowed to have such files in their repository but still ignoring generated project files.

theotherjimmy:
Resolves #490

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -79,7 +79,7 @@ ignores = [
     "*.eww",
 
     # GCC make
-    "Makefile",
+    "/Makefile",
     "Debug",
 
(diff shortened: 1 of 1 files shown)


theotherjimmy:
Yes. 
from https://git-scm.com/docs/gitignore
> A leading slash matches the beginning of the pathname.

screamerbg:
LGTM

Issue #529, opened 2017-08-04
compile command generates build target directories case-sensitively


JammuKekkonen:
Although targets k64f and K64F are effectively the same, typing the targets in different cases causes new build target folder to be created and thus causes the recompilation of the whole project. Same applies to toolchain directory under the target directory.

korjaa:
Closed #530 as there are non-uppercase targets also.

korjaa:
@theotherjimmy, sorry didn't fully catch what you meant with the Mbed OS tools. I'm not familiar with the relation of mbed-cli with Mbed OS tools.

Was the idea that the platform name would be verified before compilation, but not touched in the mbed-cli phase?

theotherjimmy:
> I'm not familiar with the relation of mbed-cli with Mbed OS tools.

Simple Explanation:
mbed CLI:
  * target detection
  * Source control
  * Running tests
mbed OS tools:
  * Compiling
  * Exporting
  * Building tests

> Was the idea that the platform name would be verified before compilation, but not touched in the mbed-cli phase?

Actually, I was recommending that you just uppercase them without concern. The mbed OS tools will figure out what to build if you use the wrong case, so rely on that.

korjaa:
Thanks for that task split explanation.

About the uppercasing, wasn't that what the original PR was doing, just uppercasing without concern? Or is it that it was on the wrong place and should be at Mbed OS tools.

theotherjimmy:
For anyone that would like to help out: I think that upcasing the _directory names_ would be a path forward here. I don't currently have time, and I would like to see this changed.

screamerbg:
This is now available in Mbed CLI 1.7.3

PR #530, opened 2017-08-04
Added automatic uppercase conversion to --toolchain and --target.

korjaa:
Any comments on this?

related to #529

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -2158,8 +2158,8 @@ def status_(ignore=False):
 
 # Compile command which invokes the mbed OS native build system
 @subcommand('compile',
-    dict(name=['-t', '--toolchain'], help='Compile toolchain. Example: ARM, GCC_ARM, IAR'),
-    dict(name=['-m', '--target'], help='Compile target MCU. Example: K64F, NUCLEO_F401RE, NRF51822...'),
+    dict(name=['-t', '--toolchain'], help='Compile toolchain. Example: ARM, GCC_ARM, IAR', type=lambda s: s.upper()),
+    dict(name=['-m', '--target'], help='Compile target MCU. Example: K64F, NUCLEO_F401RE, NRF51822...', type=lambda s: s.upper()),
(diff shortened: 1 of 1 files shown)


JammuKekkonen:
PR seems to work as expected; both toolchain and target directories are now always uppercased.

screamerbg:
@korjaa @JammuKekkonen Not sure that this will work as not all targets are upper case https://github.com/ARMmbed/mbed-os/blob/master/targets/targets.json

A possible solution is for the target folder parameter to be upper-cased when passed to the build tools (but not the target name), not on input.

korjaa:
Aa, definitely true. Did cross my mind, but didn't know where to look for all targets.

korjaa:
For me, it feels that the most logical way would be to make the target case sensitive, so that typing "k64F" would cause an error as it's not "K64F"

screamerbg:
@theotherjimmy See above.

bridadan:
+1 for enforcing consistent casing! I'd probably go with upper case for legacy reasons.

theotherjimmy:
@korjaa You don't have to worry about the case of targets, as the mbed OS tools will convert as appropriate.

@screamerbg We _could_ make it an error, but with the "you got the case wrong" message. I'm not willing to make something that worked before stop working with no help to the user. Personally, I really like not having to use caps lock or the shift key when doing this stuff, but my preference may not be universal. I know that @sg- shares it though.

On that note, @korjaa You might want to drop this PR and propose something like that to the mbed OS tools. I think it should just be as simple as dropping the "force" from the type function. Let me know if you want assistance..

korjaa:
@theotherjimmy Yes, the compilation works with both and that's the issue. It accepts the platform in anycase and then under `BUILD/` it creates a folder using that anycase name. Now, if you have a window open where your drag and drop your firmware, you could be using the previous build with a correct name from `BUILD/K64F` instead of the typoed build from `BUILD/k64F`. Or if you are using alias like below:
```
$ type mbed_flash 
mbed_flash is aliased to `cp BUILD/K64F/GCC_ARM/*.bin /media/jaakor01/DAPLINK/ && sync'
```.

Yes, I will close this PR as the uppercase doesn't really work with the non-uppercase targets.

PR #531, opened 2017-08-04
Update link in README.md

AnotherButler:
Fix broken link by using long link form because the shortened form isn't working.


screamerbg:
LGTM

PR #533, opened 2017-08-07
Fix non HTTPS links to avoid security warning

Nodraak:
Firefox (and probably other browsers) display a security warning: "Parts of this page are not secure (such as images)."

Note: the patch is intended to fix [this page](https://docs.mbed.com/docs/mbed-os-handbook/en/5.5/dev_tools/cli/)


screamerbg:
Thanks

PR #555, opened 2017-10-30
Fall back to python when exec name starts with mbed

janjongboom:
I'm compiling mbed CLI through `pyinstaller`, which causes the python path to be set incorrectly.

@screamerbg @arekzaluski


theotherjimmy:
Everything looks unchanged (still working) from the pip side of things.

PR #556, opened 2017-10-30
Allow using .exe files for mbed OS tools, instead of Python

janjongboom:
This is just an idea on getting Mbed CLI to work on a non-Python computer... It requires compilation of the tools in mbed-os/tools, which is annoying as it makes Mbed OS a lot larger. Opening the PR to fuel discussion.

@theotherjimmy @arekzaluski

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -45,8 +45,10 @@ git_cmd = 'git'
 
 # override python command when running standalone Mbed CLI
 python_cmd = sys.executable
+is_standalone = False
 if os.path.basename(python_cmd).startswith('mbed'):
     python_cmd = 'python'
+    is_standalone = True
 
 ignores = [
     # Version control folders
@@ -179,7 +181,6 @@ class ProcessException(Exception):
 
 def popen(command, stdin=None, **kwargs):
     # print for debugging
-    info('Exec "'+' '.join(command)+'" in '+os.getcwd())
     try:
         proc = subprocess.Popen(command, **kwargs)
     except OSError as e:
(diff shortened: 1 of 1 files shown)


bremoran:
Can this be made to work on Mac OS X and Linux as well?

janjongboom:
@bremoran Yeah, at least on macOS. pyinstaller works fine there. For Ubuntu I'm running into issues with libusb errors.

theotherjimmy:
I probably won't get time to review this until Thursday.

mbed/mbed.py
```diff
@@ -1294,6 +1295,14 @@ def get_tools_dir(self):
 
         return self._find_file_paths(paths, 'make.py')
 
+    def get_tools_command(self, script):
+        if os.name != 'nt' \
```
screamerbg: Could you please reverse the logic here, so standalone is the special case and the default is the .py case?

arekzaluski:
I'll also try to find some time this week to test it within pyInstaller in Windows. I'll also talk with Eclipse IDE team because I know that they are interested in using pyInstaller to install Mbed CLI.

mbed/mbed.py
screamerbg: Bump @janjongboom

janjongboom:
@screamerbg Can we let this rest for a bit, until I've talked with tools team? Maybe a different approach is better.

@arekzaluski Did you talk to Eclipse IDE team? If so, what did they say?

arekzaluski:
@janjongboom Eclipse team took different approach. They are using mbed.py directly and are shipping it isolated using https://conda.io/miniconda.html. I'll be creating an installer for Mbed Studio and all required tools this sprint.

janjongboom:
@arekzaluski Does that allow installation without an internet connection? If so, that'd be a good way for the CLI installer...

arekzaluski:
Yes, it does. It supports --offline mode. It's quite likely that I'll use it also in Mbed CLI for Windows installer if everything goes smoothly for Mbed Studio.

andrewleech:
Another option to throw into the mix is briefcase:
https://github.com/pybee/briefcase

PyInstaller often runs into support issues, with packages not working properly once built into an exe. It could really add some extra support headaches to the mix.

Briefcase provides a means of packaging your python app (mped-cli) along with normal python executable and all dependencies into a normal installer (on windows) or a proper app bundle (on mac). Linux still uses the system python but it packages up all the dependencies to make it easier to distribute.

janjongboom:
@BlackstoneEngineering ^ Interesting to look at for the MacOS installer?

theotherjimmy:
@janjongboom see https://github.com/armmbed/mbed-os-tools 😄

mark-edgeworth:
We are likely to eliminate the whole tool-calling-tool strategy soon, so this is no longer relevant.

Issue #561, opened 2017-11-11
Mbed CLI should do shallow clone by default on import


janjongboom:
On slow connections (and on some fast ones too), importing an Mbed OS program using CLI is extremely slow and might take up to half an hour. Enabling cache helps, but not the first time - so bad out of box experience -, plus users forget it.

This is mainly due to the time it costs to clone mbed-os. On the super-fast Seoul internet here (110 mbps down) it takes **3 minutes, 21 seconds** to clone mbed-os. However, 99% of the users do not need the full history, they just want to build and are probably not going to contribute back to the project. So we could speed this up significantly if we'd shallow clone the repo when importing. So I'd suggest to use `--depth=1` (I see it's already in CLI) by default.

Shallow import on the same internet takes me 36 seconds.

cc @toyowata @neil-tan

janjongboom:
Ah, dupe of #560.

theotherjimmy:
Closing this issue in favor of #560

Issue #560, opened 2017-11-09
Speed up `mbed-cli new` command by performing shallow clone of mbed-os


cmonr:
When running `mbed-cli new <name>`, the entire repo history of mbed-os is cloned into the user's repository.

At the current size of the mbed-os repo, the `git clone <mbed-os repo>` command takes ~3 minutes on my system.
Replacing this with `git clone -b latest --depth=1 <mbed-os repo>` cuts this time down to ~1.5 minutes.

teetak01:
Maybe instead of being an default option, there could be an optional configurational parameter to deploy Mbed OS as shallow. This way it would affect equally `import`, `deploy `and `new`.

bridadan:
@cmonr You can actually already do this with the option `--depth [DEPTH]`. It's currently present in `mbed import`, `mbed add`, `mbed deploy`, and `mbed update`.

I don't think we can make this the default option like @teetak01 mentioned because the `.lib` files that are then turned into repositories can reference any commit in the repository's history. If the depth is set too shallow, this commit won't be present and it will error.

cmonr:
@bridadan Using `--depth` on its own is definitely not enough and can cause the entire clone to fail, but (from what I can tell) all that goes into `mbed-os.lib` is the git URL with the specific commit hash. You can still get the same hash if `-b latest` is used, with the benefit of a faster clone and less disk space used.

bridadan:
Ah I see, so you're saying just for the `mbed new` command (which always clones from the `latest` tag), just run `git clone https://github.com/ARMmbed/mbed-os -b latest` since that will be faster?

Seems fine to me, do you need some pointers on where to make the change? This is the relevant section of the CLI: https://github.com/ARMmbed/mbed-cli/blob/master/mbed/mbed.py#L1727. You'll probably need to look into the `Git` abstractions as well: https://github.com/ARMmbed/mbed-cli/blob/master/mbed/mbed.py#L610

cmonr:
Yes, and no. More specifically, right now the `mbed new` command does a clone of the entire repo, _then_ switches to the `latest` branch.

Both `--depth=1` and `-b latest` are needed, the former to speed up the clone, and the latter to make sure that all files needed are actually pulled.

bridadan:
Ah, I see, thanks for the explanation. I think that should be fine, but I think the final call may end up being @screamerbg's.

[...]

PR #565, opened 2017-11-22
Properly performs shallow clone when `depth` is used

cmonr:
When `depth` flag is used, `clone` will attempt to pull only a single commit from repos.

If a SHA in a .lib file does not match a repo ref, the `clone` will revert to the default behavior of pulling the entire repo and checking out the specific SHA.

References: 
https://github.com/ARMmbed/mbed-cli/issues/560
https://github.com/ARMmbed/mbed-cli/issues/561

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -607,21 +607,24 @@ class Git(object):
         for branch in branches: # delete all local branches so the new repo clone is not poluted
             pquery([git_cmd, 'branch', '-D', branch])
 
-    def clone(url, name=None, depth=None, protocol=None):
-        if depth:
+    def clone(url, path, rev=None, depth=None, protocol=None, name=None):
+        result = pquery([git_cmd, "ls-remote", url, (rev if rev else "HEAD")])
+        
+        if result:
             repo_name = url.split('/')[-1]
             if '.git' in repo_name:
                 repo_name = repo_name[:-4]
 
             os.mkdir(repo_name)
-            
+
             with cd(repo_name):
                 Git.init()
-                Git.fetch(url=url, branch='latest', depth=depth)
+                Git.fetch(url=url, rev=rev, depth=depth)
                 Git.checkout('FETCH_HEAD')
                 popen([git_cmd, 'remote', 'add', 'origin', url])
+            
         else:
-            popen([git_cmd, 'clone', formaturl(url, protocol), name] + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
+            popen([git_cmd, 'clone', formaturl(url, protocol), path] + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
 
     def add(dest):
         info("Adding reference "+dest)
@@ -655,9 +658,9 @@ class Git(object):
                 if not branch:
                     error(err+"Working set is not on a branch.", 1)
 
-    def fetch(url=None, branch=None, depth=None):
+    def fetch(url=None, rev=None, depth=None):
         info("Fetching revisions from remote repository to \"%s\"" % os.path.basename(os.getcwd()))
-        popen([git_cmd, 'fetch', '--tags'] + ([url] if url else []) + ([branch] if branch else ['--all']) + (['--depth', depth] if depth else []) + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
+        popen([git_cmd, 'fetch', '--tags'] + ([url] if url else []) + ([rev] if rev else ['--all']) + (['--depth', depth] if depth else []) + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
 
     def discard(clean_files=False):
         info("Discarding local changes in \"%s\"" % os.path.basename(os.getcwd()))
@@ -1111,7 +1114,7 @@ class Repo(object):
             # Main clone routine if the clone with cache ref failed (might occur if cache ref is dirty)
             if main:
                 try:
-                    scm.clone(url, path, depth=depth, protocol=protocol, **kwargs)
+                    scm.clone(url, path, rev=rev, depth=depth, protocol=protocol, **kwargs)
                 except ProcessException:
                     if os.path.isdir(path):
                         rmtree_readonly(path)
@@ -1816,18 +1819,17 @@ def import_(url, path=None, ignore=False, depth=None, protocol=None, top=True):
         else:
             error(err, 1)
    
-    action("Syncing repo.");
-    #repo.sync()
+    repo.sync()
 
     if top: # This helps sub-commands to display relative paths to the imported program
         cwd_root = repo.path
 
-    #with cd(repo.path):
-        #deploy(ignore=ignore, depth=depth, protocol=protocol, top=False)
+    with cd(repo.path):
+        deploy(ignore=ignore, depth=depth, protocol=protocol, top=False)
 
     if top:
         action("Post action.");
-        #Program(repo.path).post_action()
+        Program(repo.path).post_action()
 
 
 # Add library command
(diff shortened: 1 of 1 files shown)


cmonr:
Note, that repo checking out could be made much faster with the omission of `.postAction()` and `.sync()` calls (see first commit), but determining if those are safe for modification are our of scope for this PR.

cmonr:
Testing results on my system. Measurements were made with `time` command in Ubuntu 16.04.3 LTS VM.

Using mbed-cli v1.2.2, `mbed new ./` took _4m40.304s_.
Using modified mbed-cli, `mbed new ./` took _2m38.188s_.

mbed/mbed.py
```diff
@@ -1724,11 +1744,12 @@ def new(name, scm='git', program=False, library=False, mbedlib=False, create_onl
         p.path = cwd_root
         p.set_root()
         if not create_only and not p.get_os_dir() and not p.get_mbedlib_dir():
-            url = mbed_lib_url if mbedlib else mbed_os_url+'#latest'
+            url = mbed_lib_url if mbedlib else mbed_os_url+"#latest"
```
theotherjimmy: Is this modification required?

mbed/mbed.py
```diff
@@ -1724,11 +1744,12 @@ def new(name, scm='git', program=False, library=False, mbedlib=False, create_onl
...
             d = 'mbed' if mbedlib else 'mbed-os'
             try:
                 with cd(d_path):
                     add(url, depth=depth, protocol=protocol, top=False)
+
```
theotherjimmy: Can we drop the formatting changes?

mbed/mbed.py
```diff
@@ -1800,7 +1821,7 @@ def import_(url, path=None, ignore=False, depth=None, protocol=None, top=True):
             warning(err)
         else:
             error(err, 1)
-
+
```
theotherjimmy: Please don't add superfluous white space.

mbed/mbed.py
```diff
@@ -1836,8 +1857,8 @@ def add(url, path=None, ignore=False, depth=None, protocol=None, top=True):
     lib.write()
     repo.add(lib.lib)
 
-    if top:
-        Program(repo.path).post_action()
+    #if top:
+    #    Program(repo.path).post_action()
```
theotherjimmy: Did you mean to check in this change?

mbed/mbed.py
cmonr: Nope. Reverted.

mbed/mbed.py
cmonr: Reverted.

mbed/mbed.py
cmonr: Nope. Reverted.

mbed/mbed.py
```diff
@@ -607,8 +607,24 @@ def cleanup():
         for branch in branches: # delete all local branches so the new repo clone is not poluted
             pquery([git_cmd, 'branch', '-D', branch])
 
-    def clone(url, name=None, depth=None, protocol=None):
-        popen([git_cmd, 'clone', formaturl(url, protocol), name] + (['--depth', depth] if depth else []) + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
+    def clone(url, path, rev=None, depth=None, protocol=None, name=None):
+        result = pquery([git_cmd, "ls-remote", url, (rev if rev else "HEAD")])
+        
+        if result and rev:
+            repo_name = url.split('/')[-1]
```
screamerbg: This breaks the behavior of clone() regarding param path. As an example if path="/tmp/sometempdir", the first part of this "if" will ignore the path value and create it's own dir based on the url. The expected behavior (as show in the else) is that path is passed to git and therefore the cloned repository will be stored in path, not in a folder generated from the url name.
E.g. imagine that there's no correlation between the library folder name and the repo url, e.g. https://github.com/ARMmbed/spiflash-stm32f429-driver is referenced from spiflash-driver.lib.

mbed/mbed.py
```diff
@@ -607,8 +607,24 @@ def cleanup():
...
+            if '.git' in repo_name:
+                repo_name = repo_name[:-4]
+
+            os.mkdir(repo_name)
```
screamerbg: try statement here will help prevent ugly traceback or silent behavior (mbed CLI by default supresses tracebacks)

mbed/mbed.py
```diff
@@ -607,8 +607,24 @@ def cleanup():
...
+
+            with cd(repo_name):
+                Git.init()
```
screamerbg: Why not try to find an alternative approach instead of hacking it yourself? E.g.
1. Clone using the original depth command
2. Add an if statement if --depth is used to call fetch with revision/tag parameter.

Worst case scenario this would fetch the latest from master + latest tag, vs 2 calls git ls-remote + git fetch.

mbed/mbed.py
```diff
@@ -642,9 +658,12 @@ def publish(all_refs=None):
                 if not branch:
                     error(err+"Working set is not on a branch.", 1)
 
-    def fetch():
+    def fetch(url=None, rev=None, depth=None):
         info("Fetching revisions from remote repository to \"%s\"" % os.path.basename(os.getcwd()))
-        popen([git_cmd, 'fetch', '--all', '--tags'] + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
+        if url:
+            popen([git_cmd, 'fetch', '--tags'] + ([url] if url else []) + ([rev] if rev else []) + (["--depth", depth] if depth else []) + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
```
screamerbg: URL should be passed to format(url, protocol) for behavior consistency. See line 627 `popen([git_cmd, 'clone', formaturl(url, protocol), path] + (['-v'] if very_verbose else ([] if verbose else ['-q'])))`

mbed/mbed.py
cmonr: Ah, that's a good point. It didn't occur to me that the paths could also point to local repos.

mbed/mbed.py
cmonr: This actually is the alternative approach. The entire goal of the feature was to download as little information as possible, to speed up the command.

To use `git clone` first would defeat the purpose of the enhancement, since downloading the entire repo history is excessive if the user only wants the files from a single release. I didn't find a flag for `git clone` that would allow for cloning a repo at a specific ref, hence the split.

cmonr:
@screamerbg, thank you for listing out the suggested design patterns. Testing local git repos did pass my checks, but isn't this the kind of thing that pytest should have caught?

To your third point, I was under the impression that HG already didn't support the `--depth` parameter, which is why I thought the flag wasn't following the design pattern to begin with. My experience with HG is minimal, but this was my reference into that question when I was developing the feature: https://stackoverflow.com/questions/7934031/what-is-the-status-of-the-mercurial-shallow-clone-extension

I'll definitely look at making sure it remains compatible with local repos, and check its behavior when using the cache feature. is there an eta on v1.3? This is the first that I've heard of a roadmap for mbed-cli.

screamerbg:
@cmonr, pytest currently doesn't cover the caching feature, nor the circle CI tests. But contributions are (always) welcome :)

I haven't suggested that HG supports shallow clones - it doesn't, not in the sense of a standard HG feature. You can emulate it with a non-standard additional plugin, but it's a messy business.

Regarding the roadmap, mbed CLI is part of the Mbed OS core tools (despite being a separate repository) and as such it's part of the mbed OS roadmap. As for it's specific technical features, feel free to contact @sg- or myself if you have questions. Caching has been around as an optional (experimental) feature for about a year now. We're looking for ways to introduce this as a standard default-on feature, but stability is a concern and we've been continuously improving how caching works. The next big step for caching will be to introduce a standard interface that allows a user to manipulate their cache, e.g.
* `mbed cache [default|on|off|none|/path]`
* `mbed cache list`
* `mbed cache size`
* `mbed cache purge|clean`

Issue #557, opened 2017-10-31
mbed targets --supported is broken


BlackstoneEngineering:
TLDR:
the command `mbed target --supported` fails, but `mbed compiler --supported` and `mbed compiler -S` works. This command documented in videos and tutorials and needs to work.

Long Version:
On mbed CLI v1.2.0 the `mbed target --supported` command works, but on v1.2.2 it does not. Something changed here that broke the command. 

The output failure message is `[mbed] ERROR: Unknown Error: sequence item 4: expected string, bool found`, which looks like a python issue, but not sure. 

This was found on windows 10, using cmd. 

```shell
C:\Users\ausbla01\mbed\mbed-os>mbed target --supported
[mbed] WARNING: Could not find mbed program in current path "C:\Users\ausbla01\mbed\mbed-os".
[mbed] WARNING: You can fix this by calling "mbed new ." in the root of your program.
---
+----------------------+-----------+-----------+-----------+-----------+-----------+-----------+
| Target               | mbed OS 2 | mbed OS 5 |    ARM    |  GCC_ARM  |    IAR    |   ARMC6   |
+----------------------+-----------+-----------+-----------+-----------+-----------+-----------+
| ARCH_PRO             | Supported | Supported | Supported | Supported | Supported | Supported |
| ARM_BEETLE_SOC       | Supported | Supported | Supported | Supported | Supported | Supported |
| ARM_CM3DS_MPS2       | Supported | Supported | Supported | Supported | Supported | Supported |
| B96B_F446VE          | Supported | Supported | Supported | Supported | Supported | Supported |
| DELTA_DFBM_NQ620     | Supported | Supported | Supported | Supported | Supported | Supported |
| DISCO_F413ZH         | Supported | Supported | Supported | Supported | Supported | Supported |
| DISCO_F429ZI         | Supported | Supported | Supported | Supported | Supported | Supported |
| DISCO_F469NI         | Supported | Supported | Supported | Supported | Supported | Supported |
| DISCO_F746NG         | Supported | Supported | Supported | Supported | Supported | Supported |
| DISCO_F769NI         | Supported | Supported | Supported | Supported | Supported | Supported |
| DISCO_L072CZ_LRWAN1  | Supported | Supported | Supported | Supported | Supported | Supported |
| DISCO_L475VG_IOT01A  | Supported | Supported | Supported | Supported | Supported | Supported |
| DISCO_L476VG         | Supported | Supported | Supported | Supported | Supported | Supported |
[...]

PR #460, opened 2017-03-22
Supported lists

mbartling:
Added ability to list toolchains, ides, targets in addition to the supported matrix output. Solves [#306](https://github.com/ARMmbed/mbed-cli/issues/306). 

NOTE: This depends on https://github.com/ARMmbed/mbed-os/pull/3993. @screamerbg @theotherjimmy can you guys think of a workaround? Should I just add a try catch?
```
   if supported:
        popen(['python', '-u', os.path.join(tools_dir, 'project.py')]
# The offending line
              + (['-S', supported] if supported else []) + (['-v'] if very_verbose else []),
              env=env)
        return
```

PR #572, opened 2017-12-01
Restored `mbed target --supported` behavior

cmonr:
Issue reference: https://github.com/ARMmbed/mbed-cli/issues/557

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -17,6 +17,56 @@ test:
         - cd .tests/bld-test/mbed && mbed update 85 --clean
         - cd .tests/bld-test && mbed update --clean
         - cd .tests/bld-test && mbed compile -m LPC1768 -j 0
+        - cd .tests && mbed new supported-tests
+        - |-
+            cd .tests/supported-tests
+            echo "Testing mbed-os-5.6.0"
+            mbed update --clean "mbed-os-5.6.0"
+            mbed compile -S
+            mbed toolchain -S
+            mbed target -S
+            mbed export -S
+        - |-
+            cd .tests/supported-tests
+            echo "Testing mbed-os-5.5.0"
+            mbed update --clean "mbed-os-5.5.0"
+            mbed compile -S
+            mbed toolchain -S
+            mbed target -S
+            mbed export -S
+        - |-
+            cd .tests/supported-tests
+            echo "Testing mbed-os-5.4.0"
+            mbed update --clean "mbed-os-5.4.0"
+            mbed compile -S
+            mbed toolchain -S
+            mbed target -S
+            mbed export -S 
+        - |-
+            cd .tests/supported-tests
+            echo "Testing mbed-os-5.3.0"
+            mbed update --clean "mbed-os-5.3.0"
+            mbed compile -S
+            mbed toolchain -S
+            mbed target -S
+            mbed export -S
+        - |-
+            cd .tests/supported-tests
+            echo "Testing mbed-os-5.2.0"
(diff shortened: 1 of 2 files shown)


screamerbg:
The underlying problem is the introduction of the `matrix` parameter, which breaks for older releases of Mbed OS (5.1-5.5.7), basically all releases since July 2016 til July 2017. See #575.

This PR doesn't fix it (rather makes it worse). Appropriate fix should revert the `matrix` behavior introduced in #460 which breaks backwards compatiblity. Notice that make.py doesn't require the `matrix` param in mbed-os-5.5, therefore this looks like a safe approach

```
/tmp/1/test-program:$ mbed update mbed-os-5.5.7
/tmp/1/test-program:$ /usr/bin/python -u /private/tmp/1/test-program/mbed-os/tools/make.py -S 
+----------------------+-----------+-----------+-----------+-----------+-----------+
| Target               | mbed OS 2 | mbed OS 5 |    ARM    |  GCC_ARM  |    IAR    |
+----------------------+-----------+-----------+-----------+-----------+-----------+
| ARCH_PRO             | Supported | Supported | Supported | Supported | Supported |
+----------------------+-----------+-----------+-----------+-----------+-----------+

[snip]

Supported targets: 92

```

theotherjimmy:
@cmonr @screamerbg We may be able to solve this by passing nothing when the argument to `-S/--supported` is `"matrix"`, as this is the default, and the old behavior.

theotherjimmy:
@screamerbg Could you advise us on how to add a test that verifies that we do this in a backwards compatible way?

screamerbg:
@cmonr Regarding how to test supported, add to circle something after `mbed new` like
`mbed targets -S`
`mbed compile -S`
`mbed update mbed-os-5.5.1`
`mbed targets -S`
`mbed compile -S`
`mbed update mbed-os-5.4.1`
`mbed targets -S`
`mbed compile -S`
`mbed update mbed-os-5.3.1`
`mbed targets -S`
`mbed compile -S`
`mbed update mbed-os-5.2.1`
`mbed targets -S`
`mbed compile -S`
`mbed update mbed-os-5.1.1`
`mbed targets -S`
`mbed compile -S`

screamerbg:
@theotherjimmy Perhaps the same test can be invoked for `mbed export -S`?

cmonr:
@screamerbg Well that's amusing. Regarding how to test, that's effectively what I'm doing right now, except I sped it up by having separate directories checked out to their appropriate versions.

screamerbg:
@cmonr No pretty but it would do the job. You can make it look neat if you iterate through `mbed releases` (once it lands to master), checkout every `MINOR.1` release and run `mbed compile -S`, `mbed export -S` and `mbed target -S`

cmonr:
@screamerbg Any particular reason why `MINOR.1` instead of `MINOR.0`? Just curious.

cmonr:
Don't merge yet until tests are written (tomorrow).

Tested using `mbed compile -S`, `mbed target -S`, and `mbed toolchain -S`, and `mbed export -S` on mbed-os-5.x.0 versions, and latest version.

Looking for guidance from @mbartling on how to test for his features.

screamerbg:
@cmonr if I'm not mistaken, one of the very first versions of mbed OS had the exporters broken in the x.0 release.

screamerbg:
@cmonr Any update on this?

cmonr:
@screamerbg The tests have been added in. I didn't find any issues with exporting from 5.1.0 and up.

PR #573, opened 2017-12-02
Revert "Properly performs shallow clone when `depth` is used"

screamerbg:
Reverts ARMmbed/mbed-cli#565

Thanks for the contribution @cmonr, but unfortunately this introduces behavior changes to Git.clone() which need addressing. Please see https://github.com/ARMmbed/mbed-cli/pull/565#pullrequestreview-80675869.

I think that the majority of the issues can be addressed by:
1. Let Git handle a first successful clone with depth param.
2. Add a condition `if depth and rev:` which calls Git.fetch() with revision/tag parameters inside the already cloned repository/folder.

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -281,7 +281,7 @@ class Bld(object):
                 else:
                     shutil.rmtree(fl)
 
-    def clone(url, path=None, rev=None, depth=None, protocol=None):
+    def clone(url, path=None, depth=None, protocol=None):
         m = Bld.isurl(url)
         if not m:
             raise ProcessException(1, "Not a library build URL")
@@ -395,7 +395,7 @@ class Hg(object):
     def cleanup():
         return True
 
-    def clone(url, name=None, rev=None, depth=None, protocol=None):
+    def clone(url, name=None, depth=None, protocol=None):
         popen([hg_cmd, 'clone', formaturl(url, protocol), name] + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
 
     def add(dest):
@@ -607,24 +607,8 @@ class Git(object):
         for branch in branches: # delete all local branches so the new repo clone is not poluted
             pquery([git_cmd, 'branch', '-D', branch])
 
-    def clone(url, path, rev=None, depth=None, protocol=None, name=None):
-        result = pquery([git_cmd, "ls-remote", url, (rev if rev else "HEAD")])
-        
-        if result and rev:
-            repo_name = url.split('/')[-1]
-            if '.git' in repo_name:
-                repo_name = repo_name[:-4]
-
-            os.mkdir(repo_name)
-
-            with cd(repo_name):
-                Git.init()
-                Git.fetch(url=url, rev=rev, depth=depth)
-                Git.checkout('FETCH_HEAD')
-                popen([git_cmd, 'remote', 'add', 'origin', url])
-            
-        else:
-            popen([git_cmd, 'clone', formaturl(url, protocol), path] + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
+    def clone(url, name=None, depth=None, protocol=None):
+        popen([git_cmd, 'clone', formaturl(url, protocol), name] + (['--depth', depth] if depth else []) + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
 
     def add(dest):
         info("Adding reference "+dest)
@@ -658,12 +642,9 @@ class Git(object):
                 if not branch:
                     error(err+"Working set is not on a branch.", 1)
 
-    def fetch(url=None, rev=None, depth=None):
+    def fetch():
         info("Fetching revisions from remote repository to \"%s\"" % os.path.basename(os.getcwd()))
-        if url:
-            popen([git_cmd, 'fetch', '--tags'] + ([url] if url else []) + ([rev] if rev else []) + (["--depth", depth] if depth else []) + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
-        else:
-            popen([git_cmd, 'fetch', '--tags', '--all'] + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
+        popen([git_cmd, 'fetch', '--all', '--tags'] + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
 
     def discard(clean_files=False):
         info("Discarding local changes in \"%s\"" % os.path.basename(os.getcwd()))
@@ -1117,7 +1098,7 @@ class Repo(object):
             # Main clone routine if the clone with cache ref failed (might occur if cache ref is dirty)
             if main:
                 try:
-                    scm.clone(url, path, rev=rev, depth=depth, protocol=protocol, **kwargs)
+                    scm.clone(url, path, depth=depth, protocol=protocol, **kwargs)
                 except ProcessException:
                     if os.path.isdir(path):
                         rmtree_readonly(path)
@@ -1805,7 +1786,7 @@ def import_(url, path=None, ignore=False, depth=None, protocol=None, top=True):
         with cd(repo.path):
             Program(repo.path).set_root()
             try:
-                if repo.rev and repo.getrev() != repo.rev and not depth:
+                if repo.rev and repo.getrev() != repo.rev:
                     repo.checkout(repo.rev, True)
             except ProcessException as e:
                 err = "Unable to update \"%s\" to %s" % (repo.name, repo.revtype(repo.rev, True))


cmonr:
@screamerbg You're correct in that `git clone` no longer properly works with local directories, an oversight on my part.

I however don't agree with the assessment that the majority of the problem cam be solved by first invoking a `git clone`.

Imo, whenever a user intends to use the `--depth` parameter with a git repo, they're wanting the contents at a specific commit (branch, tag, etc...) and don't care about the history, typically to save hard drive space and speed up the clone. Performing a `git clone` first negates both of these advantages.

Second, when a user is pulling repos with an arbitrary tag (say from mbed-os.lib with the _latest_ tag), when the `git clone` command is used with `--depth`, if the repo is several commits ahead of the tag, the clone will fail.

screamerbg:
@cmonr I did some analysis about the actual difference between shallow clone and normal clone.

As you'd see from the analysis below, the actual download reduction is 23%, not in half. I've also checked the downloaded index size and came with similar percentage (24%). This is far from the originally stated double speed up.


**Shallow clone (based on your implementation):**
* 137MB download
* 141MB index
* 598MB codebase (@latest) + index

Log
```
/tmp/1:$ mkdir mbed-os-shallow
/tmp/1:$ cd mbed-os-shallow/
/tmp/1/mbed-os-shallow:$ git init
Initialized empty Git repository in /private/tmp/1/mbed-os-shallow/.git/

/tmp/1/mbed-os-shallow:$ git fetch https://github.com/ARMmbed/mbed-os --tags latest --depth 1
remote: Counting objects: 49597, done.
remote: Compressing objects: 100% (21815/21815), done.
remote: Total 49597 (delta 31145), reused 41296 (delta 25240), pack-reused 0
Receiving objects: 100% (49597/49597), 137.31 MiB | 8.23 MiB/s, done.
Resolving deltas: 100% (31145/31145), done.
[snip]

/tmp/1/mbed-os-shallow:$ git checkout FETCH_HEAD
Checking out files: 100% (11866/11866), done.
[snip]

/tmp/1/mbed-os-shallow:$ du -chs .git/
141M	.git/
141M	total

/tmp/1/mbed-os-shallow:$ du -chs .
598M	mbed-os-shallow/
598M	total
```


**Normal clone**
* 180MB download
* 186MB index
* 644MB codebase (@latest) + index

```
/tmp/1:$ git clone https://github.com/ARMmbed/mbed-os
Cloning into 'mbed-os'...
remote: Counting objects: 158654, done.
remote: Compressing objects: 100% (2/2), done.
remote: Total 158654 (delta 4), reused 3 (delta 3), pack-reused 158649
Receiving objects: 100% (158654/158654), 180.55 MiB | 8.08 MiB/s, done.
Resolving deltas: 100% (107359/107359), done.
Checking out files: 100% (12310/12310), done.

/tmp/1:$ cd mbed-os

/tmp/1/mbed-os:$ du -chs .git
186M	.git
186M	total

/tmp/1/mbed-os:$ du -chs .
644M	mbed-os
644M	total
```

The actual difference is 43MB download, 45MB index and 46MB. Also in your original PR you were stating that your patch doubles the speed of clone, but if you do the math, it only reduces the downloaded index size with 23%, not in half. Also the checked out code size is the same. Perhaps your original speedup measurement didn't add for the checkout time after the fetch command?

Also, I did some analysis for the method I suggested.

**Suggested method**
* 135MB download
* 186MB index
* 613MB codebase (@latest) + index
Basically the download size is identical, though the download time is bigger.

Log:
```
/tmp/1:$ git clone https://github.com/ARMmbed/mbed-os mbed-os-shallow-new --depth 1
Cloning into 'mbed-os-shallow-new'...
remote: Counting objects: 13334, done.
remote: Compressing objects: 100% (8290/8290), done.
remote: Total 13334 (delta 5912), reused 8520 (delta 4427), pack-reused 0
Receiving objects: 100% (13334/13334), 37.71 MiB | 6.81 MiB/s, done.
Resolving deltas: 100% (5912/5912), done.
Checking out files: 100% (12310/12310), done.

/tmp/1:$ cd mbed-os-shallow-new/
/tmp/1/mbed-os-shallow-new:$ git fetch https://github.com/ARMmbed/mbed-os --tags latest --depth 1
remote: Counting objects: 36266, done.
remote: Compressing objects: 100% (17319/17319), done.
remote: Total 36266 (delta 25221), reused 26607 (delta 17022), pack-reused 0
Receiving objects: 100% (36266/36266), 97.41 MiB | 8.14 MiB/s, done.
Resolving deltas: 100% (25221/25221), completed with 2328 local objects.
From https://github.com/ARMmbed/mbed-os

/tmp/1/mbed-os-shallow-new:$ du -chs .git
156M	.git
156M	total

/tmp/1/mbed-os-shallow-new:$ du -chs .
613M	mbed-os-shallow-new/
613M	total
```

Last but not least, if the goal is to provide an absolute minimal download index size, then --tags should be omitted.

**Minimalistic approach**
* 38MB download
* 53MB index
* 511MB codebase (@latest) + index
Basically the download size is identical, though the download time is bigger.
Log:
```
/tmp:/1:$ git clone https://github.com/ARMmbed/mbed-os mbed-os-shallow-mini --depth 1
Cloning into 'mbed-os-shallow-mini'...
remote: Counting objects: 13334, done.
remote: Compressing objects: 100% (8290/8290), done.
remote: Total 13334 (delta 5912), reused 8520 (delta 4427), pack-reused 0
Receiving objects: 100% (13334/13334), 37.71 MiB | 6.99 MiB/s, done.
Resolving deltas: 100% (5912/5912), done.
Checking out files: 100% (12310/12310), done.

/tmp/1:$ cd mbed-os-shallow-mini/
/tmp/1/mbed-os-shallow-mini:$ git fetch https://github.com/ARMmbed/mbed-os latest --depth 1
remote: Counting objects: 885, done.
remote: Compressing objects: 100% (722/722), done.
remote: Total 885 (delta 536), reused 347 (delta 77), pack-reused 0
Receiving objects: 100% (885/885), 10.94 MiB | 2.57 MiB/s, done.
Resolving deltas: 100% (536/536), completed with 432 local objects.
From https://github.com/ARMmbed/mbed-os
 * tag               latest     -> FETCH_HEAD

/tmp/1/mbed-os-shallow-mini:$ du -chs .git
 53M	.git
 53M	total

/tmp/1/mbed-os-shallow-mini:$ du -chs .
511M	.
511M	total
```

If you dig deeper into Git, you might find that when you fetch with `git fetch origin --tags latest --depth 1` you're essentially creating a shallow clone for all referenced objects / tags, not just that single tag you're interested in. This doesn't deliver much savings, not double at least. You're only saving from commits / branches that aren't referenced as tags, which in this case is 23% of the index, but prevents the user from updating to master or feature branches. If the true goal is to deliver the absolute minimum index, then method 4 is the way to go.

**"mbed new ." Speed Comparison
I've also compared the download times of the mbed CLI 1.2.2 vs master containing your patch:

```
/tmp/1:$ time mbed new test-full
[mbed] Creating new program "test-full" (git)
[mbed] Adding library "mbed-os" from "https://github.com/ARMmbed/mbed-os" at branch latest
[mbed] Updating reference "mbed-os" -> "https://github.com/ARMmbed/mbed-os/#182bbd51bc8d1b6217477c2b5de3f1d0c5e6249f"

real	0m36.677s
user	0m11.987s
sys	0m6.396s
```

```
/tmp/1:$ time python ~/Work/mbed-cli/mbed/mbed.py new test-shallow --depth 1
[mbed] Creating new program "test-shallow" (git)
[mbed] Adding library "mbed-os" from "https://github.com/ARMmbed/mbed-os" at branch latest
[mbed] Updating reference "mbed-os" -> "https://github.com/ARMmbed/mbed-os/#182bbd51bc8d1b6217477c2b5de3f1d0c5e6249f"

real	0m34.775s
user	0m9.030s
sys	0m4.725s
```

I hope you'd agree that 2-3 seconds is insignificant difference.


In conclusion, regardless of the best approach, e.g. 1) absolute minimum index as described in method 4; or 2) extended index with all tags, the original PR requires more work. As I suggested in the revert description, this PR should also make use the functionality of the SCM as much as possible, not hack it around (mkdir + init + fetch + add url with no error handling, nor new behaviors etc).

cmonr:
@screamerbg Thank you for the in-depth response. I re-ran benchmarks in my VM and added network throttling with the `trickle` command to eliminate network state as a variable.
I also ran tests on my personal linux machine since I suspected that the incredibly large _sys_ times were due to the VM.

**Testing results**
Ubuntu 16.04 VM:
`$ time trickle -s -d 2048 mbed new . -vv`

> real 6m38.885s
> user 0m16.228s
> sys 1m39.584s

`$ time trickle -s -d 2048 mbed new . --depth=1 -vv`

> real 4m17.812s
> user 0m11.076s
> sys 0m51.396s

Arch Linux 2017.12.01:
`$ /usr/bin/time -p -o results1 trickle -s -d 2048 mbed new . -vv`

> real 98.48
> user 24.89
> sys 6.00

`$ /usr/bin/time -p -o results2 trickle -s -d 2048 mbed new . --depth=1 -vv`

> real 83.99
> user 17.21
> sys 4.90

Done on a bare-metal machine, and not a VM, I completely agree that the speed improvements are considerable much smaller. However, this is assuming that all users have access to a network with reasonable bandwidth, which isn't always the case. 
Caching is only a solution _after_ the initial mbed-os download, can still lead to a poor out-of-box experience.

Additionally, I'm not quite sure where the references to 2x the improvement are coming from. Best that I can tell, you're referring to [this](https://github.com/ARMmbed/mbed-cli/pull/565) PR, in which I started off claiming that the PR shaved three minutes off of a 5 minute operation, but whose estimates were then updated with comments following those commits. 
Or was it referring to [this](https://github.com/ARMmbed/mbed-cli/issues/560) issue request, where I only took into account the git command, and not the entire `mbed new` command?

**Overall goal of the enhancement**
It seems like we're talking about two different goals, so to make myself a bit clearer, what I was aiming to do with this enhancement was to speed up the _initial_ mbed-os clone from the repo (github in this case) by _only_ pulling the files needed to compile the library at the _latest_ tag. I also tried to be as noninvasive to the CLI and inner workings of the Repo as possible (which didn't pan out due to omitting local paths).

I know that whenever I use the depth parameter when cloning a git repo, my intention is to only download what I need in order to use the repo. If I actually want to do work _in_ the repo, I'll download a much larger chunk of the repo, if not all of it. It was this user-interaction goal that guided me to the particular implementation that I ended up with.

**Comments on your tests**
Shallow clone vs Normal clone:
- I'm not expecting the file sizes to be that big of a difference. I would expect to see improvements in download time, in particular, if the network is not good.
Suggested method
- Doing a `git clone ... --depth=1` makes the assumption that the `latest` tag is always at `HEAD`, which can't be the case if we have a tag specifically dedicated for the latest release.
Minimalistic approach (+ @theotherjimmy )
- Doing only a fetch doesn't actually pull the files from the remote, hence the `git checkout FETCH_HEAD`

**Final notes**
I agree that this enhancement needs work since it cannot handle local paths. If it's easier to revert the commit instead of submitting another PR for the improvements, then so be it. Given the goals that I had with this particular enhancement, I don't think there is a better way to get the SCM (git) to handle the particular challenges of _only_ checking out mbed-os/repos at a _specific_ tag, with _only_ the files in the repo at that point in time, and _no_ history, which is how I interpret the `--depth` flag should be used.

On the benchmarking, while I do see your point in that performing an `mbed new` on a bare-metal machine is much faster than a VM, I don't think we should ignore VMs as a platform. Caching the repo locally helps with every instance after the first where the user needs the repo in question, but this enhancement was/is focused on the first repo clone.

PR #582, opened 2017-12-08
Update CircleCI to run on Ubuntu 14.04 LTS instead of Ubutnu 12.04 LTS

cmonr:
Noticed when working on a past PR that I had to manually setup CircleCI to use Ubutnu 12.04 LTS, which is no longer being supported (see [link](https://circleci.com/blog/ubuntu-12-04-precise-build-image-end-of-life-warning/)).

PR updates only change needed to get CircleCI builds functional on 14.04.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -83,4 +83,4 @@ dependencies:
         - mkdir .tests
         - sudo add-apt-repository -y ppa:terry.guo/gcc-arm-embedded
         - sudo apt-get -y update
-        - sudo apt-get -y install gcc-arm-none-eabi=4.9.3.2015q3-1precise1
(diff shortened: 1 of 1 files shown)


theotherjimmy:
Also worth noting: Mbed OS moved to GCC 6 at version 5.6. We should consider splitting the CI based on GCC version so that we don't accidentally catch incompatibilities that are not bugs. That can be left to another PR.

theotherjimmy:
@screamerbg Is there a reason that we need to keep Circle at 12.04?

screamerbg:
@theotherjimmy none that I can think of. LGTM

Issue #600, opened 2018-01-10
Near-total test failure on Windows


davidhyman:
`=============== 3 failed, 7 passed, 30 error in 306.61 seconds ================`
A lot of them seem similar failures - for example:
```
fatal: unable to access 'https://c//users/username/appdata/local/temp/pytest-of-username/pytest-5/test_sync_update_move_git1_0/test3/': Couldn't resolve host 'c'
```

theotherjimmy:
@davidhyman #606 should have fixed this for you.

davidhyman:
thanks @theotherjimmy , wfm

PR #606, opened 2018-01-10
Fix Git URL regex so it doesn't match Windows paths 

screamerbg:
This patch fixes the regular expression that matches Git URLs so it doesn't match Windows paths. Without this patch, the path of local (unpublished) repositories on Windows, e.g. "c:/users/path/repo.git" which only exists locally, will be identified as remote repositories. This also fixes py.test on Windows #600 

```
D:\Work\neo>py.test
============================= test session starts =============================
platform win32 -- Python 2.7.11, pytest-2.9.2, py-1.4.31, pluggy-0.3.1
rootdir: D:\Work\neo, inifile:
collected 40 items

test\add_test.py ........
test\import_test.py ....
test\ls_test.py ....
test\remove_test.py ........
test\sync_update_test.py ................

========================= 40 passed in 550.85 seconds =========================
```


theotherjimmy:
@screamerbg I'll merge this now. Could you push a new patch release of mbed-cli when you get the chance?

Issue #590, opened 2017-12-21
mbed export fails with switch "--source .."


Alex-EEE:
Hi,

I'm trying to export while also including source files in the parent directory of my mbed program.  The following fails:
```
C:\files\mbed\sandbox\framework\mbed> mbed export --source .. -i IAR
Scan: ..
Scan: FEATURE_LWIP
Scan: FEATURE_STORAGE
Traceback (most recent call last):
  File "C:\files\mbed\sandbox\framework\mbed\mbed-os\tools\project.py", line 259, in <module>
    main()
  File "C:\files\mbed\sandbox\framework\mbed\mbed-os\tools\project.py", line 254, in main
    build_profile=profile, app_config=options.app_config)
  File "C:\files\mbed\sandbox\framework\mbed\mbed-os\tools\project.py", line 96, in export
    app_config=app_config)
  File "C:\files\mbed\sandbox\framework\mbed\mbed-os\tools\export\__init__.py", line 358, in export_project
    copyfile(static_file, join(export_path, basename(static_file)))
  File "C:\Python27\lib\shutil.py", line 82, in copyfile
    with open(src, 'rb') as fsrc:
IOError: [Errno 2] No such file or directory: 'C:\\files\\mbed\\sandbox\\framework\\mbed\\mbed-os\\tools\\export\\.mbed'
```

theotherjimmy:
@Alex-EEE That's going to try to scan the current directory twice... You many want to just run the export from the directory above instead.

theotherjimmy:
That being said, I don't know why you would be getting that error, unless you deleted the `.mbed` file from the tools directory.

theotherjimmy:
@Alex-EEE You seem to have left off the mcu argument. I assume that you had the `-m` argument present, as it got passed argument parsing.

Alex-EEE:
The command works when I run it "normally" (from the root of the mbed program) with no source switch.  I was able to do mbed test -compile --source .. and pull in TESTS folders from the parent of my mbed program, so I thought maybe export would work as well.  But perhaps export will be unhappy looking at folders that are not structured as mbed programs?
-MCU: I have it set in my global properties.

theotherjimmy:
@Alex-EEE 

> folders that are not structured as mbed programs

I suppose that this is possible, but unlikely.

-------------------

The error your seeing is quite odd, It's complaining about not being able to open a particular file, which exists.

I also expect that to work.

theotherjimmy:
[...]

PR #626, opened 2018-02-14
Update mbed-cli to allow different configured path for mbed-os

daid:
In not every project structure it makes sense to have the mbed-os directory at the root of your project structure.
This change allows you to configure a custom path for the mbed-os directory in your .mbed configuration.

In our case, our project calls for multiple firmware builds. But we do not want checkout mbed-os per firmware. mbed-cli allows this with the ```--source``` parameter, however, it does require that mbed-os is in the root of the project.

With this change we can setup mbed-os in a better subfolder to allow for a better project structure:
```
extlibs/mbed-os
module/mobuleA
module/moduleB
project/project1
project/project2
```
And compile those with:
```
mbed-cli config MBED_OS_DIR extlibs/mbed-os/
mbed-cli compile -f --profile release --source project/project1 --source module --source extlibs/mbed-os --build BUILD/project1
mbed-cli compile -f --profile release --source project/project2 --source module --source extlibs/mbed-os --build BUILD/project2
```

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1384,6 +1384,8 @@ class Program(object):
 
     # Gets mbed OS dir (unified)
     def get_os_dir(self):
+        if self.get_cfg('MBED_OS_DIR', None) is not None:
+            return self.get_cfg('MBED_OS_DIR')
         if os.path.isdir(os.path.join(self.path, 'mbed-os')):
             return os.path.join(self.path, 'mbed-os')
         elif self.name == 'mbed-os':


theotherjimmy:
This is a very interesting, and quite simple, change. I'm okay with it. @screamerbg What do you think?

loverdeg-ep:
Yes please. I'd like to use this to support adding mbed-os as a git submodule versus having mbed-cli import it.

screamerbg:
@daid As @theotherjimmy pointed, it's a simple and non-breaking change. Thanks!

bmcdonnell-ionx:
Does this work with exporters?

theotherjimmy:
@bmcdonnell-ionx This should work with all mbed verbs, export included.

daid:
@bmcdonnell-ionx #590 is a bigger problem for complex project setups with the exporter. As this patch only helps the python tools find the actual exporter (and more things), I don't think it causes any problems. And any problems that it would cause are most likely bugs due to wrong assumptions on paths.

PR #634, opened 2018-03-05
Add ability to specify/save profile with configuration file

### LINKED ISSUE #633: Should be able to specify profile with configuration file
I want to enable C++11 for a project.

Right now that means adding the `-std=c++11` flag to a toolchain profile, then adding the profile flag to `mbed compile` every time it is run.

Obviously this could be solved with an alias or a Makefile, but it would be cool if the profile could be saved in the `.mbed` file, like the target and toolchain, so it is remembered and optionally could be pushed to a repo.

Alternatively, would adding something to the mbed_app.json file be better? Maybe something like `target_overrides`, but called `profile_appends` or `toolchain_appends`.

I'm very happy to put together a pull request on this, I just wanted to know if this was a feature people were interested in/happy to accept.

lfdebrux:
This pull request is to try and resolve #633 

This patch treats the profile like the target and the toolchain currently are; it has methods in the Program class for getting and setting from a Cfg, and is automatically saved to the config file as a default if set at the command line. Additionally, `mbed config profile` works as expected.

Multiple profiles are handed by allowing the value to be a list of space-separated strings.

Tests for the `mbed config profile` functionality are in `test/profile_test.py`

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1484,7 +1484,7 @@ class Program(object):
     def post_action(self):
         mbed_tools_path = self.get_tools_dir()
 
-        if not mbed_tools_path and (self.is_classic or os.path.exists(os.path.join(self.path, Cfg.file))):
+        if not mbed_tools_path and self.is_classic:
             self.add_tools(os.path.join(self.path, '.temp'))
             mbed_tools_path = self.get_tools_dir()
 
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1560,11 +1560,18 @@ class Program(object):
             error("Please specify toolchain using the -t switch or set default toolchain using command 'mbed toolchain'", 1)
         return tchain
 
-    def set_defaults(self, target=None, toolchain=None):
+    def get_profile(self, profile=None):
+        profile_cfg = self.get_cfg('PROFILE')
+        profile_cfg = profile_cfg.split() if profile_cfg else []
+        return profile if profile else profile_cfg
+
+    def set_defaults(self, target=None, toolchain=None, profile=None):
         if target and not self.get_cfg('TARGET'):
             self.set_cfg('TARGET', target)
         if toolchain and not self.get_cfg('TOOLCHAIN'):
             self.set_cfg('TOOLCHAIN', toolchain)
+        if profile and not self.get_cfg('PROFILE'):
+            self.set_cfg('PROFILE', (' ').join(profile))
 
     def get_macros(self):
         macros = []
@@ -2399,6 +2406,7 @@ def compile_(toolchain=None, target=None, profile=False, compile_library=False, 
 
     target = program.get_target(target)
     tchain = program.get_toolchain(toolchain)
+    profile = program.get_profile(profile)
     macros = program.get_macros()
 
     if compile_config:
@@ -2470,7 +2478,7 @@ def compile_(toolchain=None, target=None, profile=False, compile_library=False, 
                 if not reset_dev(detected['port']):
                     error("Unable to reset the target board connected to your system.\nThis might be caused by an old interface firmware.\nPlease check the board page for new firmware.", 1)
 
-    program.set_defaults(target=target, toolchain=tchain)
+    program.set_defaults(target=target, toolchain=tchain, profile=profile)
 
 
 # Test command
@@ -2502,6 +2510,7 @@ def test_(toolchain=None, target=None, compile_list=False, run_list=False, compi
 
     target = program.get_target(target)
     tchain = program.get_toolchain(toolchain)
+    profile = program.get_profile(profile)
     macros = program.get_macros()
     tools_dir = program.get_tools()
     build_and_run_tests = not compile_list and not run_list and not compile_only and not run_only
@@ -2572,7 +2581,7 @@ def test_(toolchain=None, target=None, compile_list=False, run_list=False, compi
                   + args,
                   env=env)
 
-    program.set_defaults(target=target, toolchain=tchain)
+    program.set_defaults(target=target, toolchain=tchain, profile=profile)
 
 
 # Export command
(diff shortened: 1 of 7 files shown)


mbed/mbed.py
```diff
@@ -2752,7 +2761,7 @@ def target_(name=None, global_cfg=False, supported=False):
     dict(name='name', nargs='?', help='Default toolchain name. Example: ARM, GCC_ARM, IAR'),
     dict(name=['-G', '--global'], dest='global_cfg', action='store_true', help='Use global settings, not local'),
     dict(name=['-S', '--supported'], dest='supported', action='store_true', help='Shows supported matrix of targets and toolchains'),
-    help='Set or get default toolchain\n\n',
+    help='Set or get default toolchain\n',
```
theotherjimmy: Could you move this unrelated change to another PR?

theotherjimmy:
@lfdebrux It looks like an unrelated failure. I'm going to restart the CI.

mbed/mbed.py
lfdebrux: I can move it to a separate PR, but I think it's related? It changes the cosmetics of the `mbed help` output, it means that target, toolchain and profile are grouped together.

Without it the output would be

        config     Tool configuration
        target     Set or get default target 
        toolchain  Set or get default toolchain
        
        profile    Set or get default profile(s)

theotherjimmy: :+1: Thanks for the explanation.

theotherjimmy:
@lfdebrux The CI failure reproduced. It's probably related then. Could you take a look?

lfdebrux:
@theotherjimmy 

I'm not sure how to fix this, but I don't _think_ it's related to this PR.

I tried running the commands from the circle.yaml file in a clean docker image. I cloned my fork from  https://github.com/lfdebrux/mbed-cli and tried the build against `master`, `1.4.0`, and `save-profile`, and got the same error with each. Looking at the [circle logs](https://circleci.com/gh/ARMmbed/mbed-cli), builds seem to have been failing from February 18th?

The error message suggests that there are some packages missing, as far as I can see the `check_requirements` method normally deals with that by using pip to install missing requirements. This leads me to believe that the issue might be that the `hg-test` repo is using a version of `mbed-os` that doesn't include `requirements.txt` files?

Not sure where to go from here.

theotherjimmy:
@lfdebrux Thanks for looking into that, and reproducing on master. @cmonr Could you take a look at this circle issue?

cmonr:
@theotherjimmy Sure thing. I'll see what I can find.

cmonr:
@lfdebrux Fyi, I'll be away starting tomorrow, and won't be back until next week. I'll see what I can find today, but can't make any promises.

screamerbg:
I'd like @sg- take on this, i.e. having the default mbed OS profile changed through config option in mbed CLI. Previously @sg- had concerns about reproducibility for when a different user doesn't have the mbed CLI specific config.

screamerbg:
@lfdebrux @cmonr @theotherjimmy CI is now fixed.

lfdebrux:
@screamerbg thanks for fixing the CI.

I see your point about reproducibility, I guess it would depend on the use case. It does save having to remember specifying the profile every time you run `mbed compile` which is a plus, but it probably isn't a good fit for my use case, where I want to make sure the project is always built with a certain profile.

theotherjimmy:
@lfdebrux You could check in the `.mbed` file. That file contains local settings.

screamerbg:
Also may be `mbed profile` could be a local only config option?

theotherjimmy:
I was under the impression that `mbed profile` was locally configurable. Sorry I read the code wrong.

screamerbg:
@lfdebrux From technical standpoint, this is a really good PR. Thanks for walking the extra mile!

lfdebrux:
@screamerbg thanks.

@theotherjimmy @screamerbg in the PR as-is, profile can be configured locally or globally. So yeah actually I could just commit the project `.mbed` file. I'm not sure whether being able to configure the profile globally is a good thing, I put it in because I wanted the functionality to be similar to `config target` and `config toolchain`, but I could take it out if having that feature is a bad idea.

theotherjimmy:
I think it would be best to have locally-configured profiles only. @screamerbg, It looks like you're of the same mind from the comment history.

@lfdebrux Could you remove the ability for a user to configure a build profile globally?

lfdebrux:
Sorry for the delay in sorting out the local-only profile, I was on holiday.

I fixed it in the simplest way I could think of, I've added tests to check the functionality but it seems to do the job.

screamerbg:
@lfdebrux Thanks for this. Can I please ask you to:
* fix the issues after the rebase, e.g. the Files Changed tab shows changes in 7 files, including version changes (from 1.4.0 to 1.5.0) etc. 
* remove the verb "mbed profile" for now and leave profile as a config option through `mbed config`. We'd like to revisit the current convention for mbed verbs (`mbed <something>`) between config options and actual functionality. There's an argument that if target, toolchain and profile are used, then we should introduce every config option as verb and as we don't have a position on this at the moment, I'd like to unblock this PR by providing the actual config functionality.

I hope that this makes sense and apologies for the delay for accepting this PR.

mbed/mbed.py
```diff
@@ -37,7 +37,7 @@
 
 
 # Application version
-ver = '1.4.0'
+ver = '1.5.0'
```
screamerbg: Seems that the rebase picked other changes as well. Please rebase correctly so this picks only the changes against the current mbed CLI master.

mbed/mbed.py
```diff
@@ -2780,6 +2792,26 @@ def toolchain_(name=None, global_cfg=False, supported=False):
         return compile_(supported=supported)
     return config_('toolchain', name, global_cfg=global_cfg)
 
+@subcommand('profile',
```
screamerbg: Looks like duplicate code, which might be result of the rebase issue I mentioned above. Do you think you could remove the subcommand for now until there's more clarity on `mbed <verbs>`

theotherjimmy:
@lfdebrux if you're still interested in this feature, but no longer have time to contribute, we can pick this up for you. Would you like this feature?

PR #640, opened 2018-03-13
Fixing code

OPpuolitaival:
Small fixing like: 
- Initialise variable before usage
- Match string arguments amount
- Remove unused variables

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -2059,7 +2059,7 @@ def publish(all_refs=None, msg=None, top=True):
     if repo.is_local:
         error(
             "%s \"%s\" in \"%s\" is a local repository.\nPlease associate it with a remote repository URL before attempting to publish.\n"
-            "Read more about publishing local repositories here:\nhttps://github.com/ARMmbed/mbed-cli/#publishing-local-program-or-library" % ("Program" if top else "Library", repo.name, repo.path, repo.scm.name), 1)
+            "Read more about publishing local repositories here:\nhttps://github.com/ARMmbed/mbed-cli/#publishing-local-program-or-library" % ("Program" if top else "Library", repo.name, repo.path), 1)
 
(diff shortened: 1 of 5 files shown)


OPpuolitaival:
@theotherjimmy I will fix that problem

mbed/mbed.py
```diff
@@ -1278,6 +1279,7 @@ def get_cache(self, url):
     def set_cache(self, url):
         up = urlparse(formaturl(url, 'https'))
         if self.cache and up and up.netloc and os.path.isdir(self.path):
+            cpath = 'none'
```
screamerbg: This is not right. `cpath = os.path.join(...` should be moved before the `try:` instead

OPpuolitaival:
Fixed @screamerbg mentioned problem and added auto formatting

screamerbg:
@OPpuolitaival LGTM!

screamerbg:
@OPpuolitaival Could you please rebase?

OPpuolitaival:
@screamerbg rebased

screamerbg:
@OPpuolitaival Can you remove the autoformatting? That's way too intrusive and I don't agree with some of the autoformatting rules.

screamerbg:
@OPpuolitaival I mean whatever was introduced in https://github.com/ARMmbed/mbed-cli/pull/640/commits/85829a210cf4b61ddad478e33c7ef2b8c4a5fa5e

OPpuolitaival:
@screamerbg autoformatting removed. I did ask from Jimmy and he was okay with those and therefore added it in here

screamerbg:
@OPpuolitaival Thanks. Current change set is 15 changed lines vs 580 with autoformatter.

OPpuolitaival:
@screamerbg Sure, that is the idea of autoformatting that it does big job automatically

PR #643, opened 2018-03-15
Check whether the cached repo scm matches the clone scm

screamerbg:
This adds an additional check whether the cached repository matches the current SCM context during Repo.clone(). This aims to prevent cases where a cached Mercurial repo is used in Git context for example.


screamerbg:
cc @theotherjimmy

Issue #639, opened 2018-03-09
[Feature Request] Serial Monitor


yennster:
I would love to see a serial monitor built in as a command line -- option within Mbed CLI. 

I am developing for the ESP32 right now and have been using the [esp-idf](https://github.com/espressif/esp-idf) development framework to do a lot of C code debugging for an [Adafruit Huzzah32 Feather](https://www.adafruit.com/product/3405) board. In addition to esp-idf being extremely easy to setup, they also have a bunch of great command line tools. One of which is a call to a serial monitor that automatically opens up the serial port to your board in the terminal. This call can also be combined with a make and flash commands to be extremely useful/powerful when debugging (`make flash monitor`).

Mbed CLI already has a `-f` option to automatically flash the program to the connected board after compiling, so I think a natural next step is to create a `-s` option ("s" for serial, or something similar) to open up a serial monitor in the terminal window after flashing. esp-idf even uses a python script to accomplish this monitor creation, so I feel something similar could work seamlessly within Mbed CLI.

Here's the links to the esp-idf implementations:
* [esp-idf Github repo](https://github.com/espressif/esp-idf)
* [esp-idf `idf_monitor.py`](https://github.com/espressif/esp-idf/blob/master/tools/idf_monitor.py)
* [esp-idf IDF Monitor documentation](https://esp-idf.readthedocs.io/en/latest/get-started/idf-monitor.html#idf-monitor)

PR #650, opened 2018-03-27
New feature: Serial Terminal

screamerbg:
Mbed CLI already has a -f option to automatically flash the program to the connected board after compiling. This PR takes the next natural next step to add serial terminal via `-s/--sterm` option ("s" for serial) to `mbed compile` and `mbed detect`.

**Workflow**
This PR extends `mbed compile` and `mbed detect` by adding `-s/--sterm` option. The option can be chained with `-f/--flash` (with `mbed compile`) or `-r/--reset` (with `mbed detect`), meaning that Mbed CLI will first apply the new firmware image, reset the MCU/target/board and then open serial terminal.

The serial terminal offers few commonly used shortcuts:
* Quit: CTRL+C or CTRL+]
* Reset: CTRL+B or CTRL+R
* Echo toggle: CTRL+E
* Terminal info: TAB or CTRL+I
* Help: Ctrl+H
* Menu: Ctrl+T (PySerial)

There are much more options behind the terminal menu (Ctrl+T).

The implementation is based on PySerial, which is one of the fundamental libraries for Mbed Greentea (`mbed test`). Therefore there are no additional dependencies.

**Example usage with `mbed compile`**

```
$ mbed compile -t GCC_ARM -m NUCLEO_F401RE -f -s
Building project mbed-os-example-blinky (NUCLEO_F401RE, GCC_ARM)
Scan: .
Scan: mbed
Scan: env
Compile [100.0%]: main.cpp
Link: mbed-os-example-blinky
Elf2Bin: mbed-os-example-blinky
+------------------+-------+-------+------+
| Module           | .text | .data | .bss |
+------------------+-------+-------+------+
| [fill]           |   102 |     7 |   11 |
| [lib]/c.a        | 27166 |  2204 |   56 |
| [lib]/gcc.a      |  3752 |     0 |    0 |
| [lib]/m.a        |    88 |     0 |    0 |
| [lib]/misc       |   236 |    16 |   28 |
| main.o           |   267 |     4 |  276 |
| mbed-os/drivers  |  1301 |     4 |  100 |
| mbed-os/hal      |  1395 |     4 |   66 |
| mbed-os/platform |  3096 |     4 |  354 |
| mbed-os/rtos     |  9114 |   168 | 5989 |
| mbed-os/targets  |  7017 |     5 |  388 |
| Subtotals        | 53534 |  2416 | 7268 |
+------------------+-------+-------+------+
Total Static RAM memory (data + bss): 9684 bytes
Total Flash memory (text + data): 55950 bytes

Image: ./BUILD/NUCLEO_F401RE/GCC_ARM/mbed-os-example-blinky.bin
[mbed] Detected "NUCLEO_F401RE" connected to "/Volumes/NODE_F401RE" and using com port "/dev/tty.usbmodem1413"
[mbed] --- Terminal on /dev/tty.usbmodem1413 - 9600,8,N,1 ---
[mbed] --- Quit: CTRL+C | Reset: CTRL+B | Echo: CTRL+E ---
[mbed] --- Info: TAB    | Help:  Ctrl+H | Menu: Ctrl+T ---
$$$07200221061A657F395EF362start
something 1
something 2
something 3
something 4
something 5
something 6
something 7
something 8
something 9
```

**Example use with `mbed detect`**
```
$ mbed detect -rs
[mbed] Detected "NUCLEO_F401RE" connected to "/Volumes/NODE_F401RE" and using com port "/dev/tty.usbmodem1413"
[mbed] --- Terminal on /dev/tty.usbmodem1413 - 9600,8,N,1 ---
[mbed] --- Quit: CTRL+C | Reset: CTRL+B | Echo: CTRL+E ---
[mbed] --- Info: TAB    | Help:  Ctrl+H | Menu: Ctrl+T ---
$$$07200221061A657F395EF362start
something 1
something 2
something 3
something 4
something 5
something 6
something 7
```

Inspired by @yennster's request (#639).

@yennster Do you think you could look into adding help for this feature in the README.md?

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -2749,7 +2749,7 @@ def detect(reset=False, sterm=False):
     with cd(program.path):
         tools_dir = program.get_tools_dir()
 
-    if tools_dir:
+    if tools_dir and not (reset or sterm):
         # Prepare environment variables
         env = program.get_env()
 
@@ -2762,7 +2762,9 @@ def detect(reset=False, sterm=False):
             if very_verbose:
                 error(str(e))
     else:
-        warning("The mbed OS tools were not found in \"%s\". \nLimited information will be shown about connected mbed targets/boards" % program.path)
+        if not tools_dir:
+            warning("The mbed OS tools were not found in \"%s\". \nLimited information will be shown about connected mbed targets/boards" % program.path)
+
         targets = program.get_detected_targets()
         if targets:
             unknown_found = False
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1752,6 +1752,103 @@ def formaturl(url, format="default"):
     return url
 
 
+def cdc(port, reset=False, sterm=False, baudrate=9600, timeout=10):
+    from serial import Serial, SerialException
+
+    def get_instance(*args, **kwargs):
+        try:
+            serial_port = Serial(*args, **kwargs)
+            serial_port.flush()
+        except Exception as e:
+            error("Unable to open serial port connection to \"%s\"" % port)
+            return False
+        return serial_port
+
+    def cdc_reset(serial_instance):
+        try:
+            serial_instance.sendBreak()
+        except:
+            try:
+                serial_instance.setBreak(False) # For Linux the following setBreak() is needed to release the reset signal on the target mcu.
+            except:
+                result = False
+
+    def cdc_term(serial_instance):
+        import serial.tools.miniterm as miniterm
+
+        term = miniterm.Miniterm(serial_instance, echo=True)
+        term.exit_character = '\x03'
+        term.menu_character = '\x14'
+        term.set_rx_encoding('UTF-8')
+        term.set_tx_encoding('UTF-8')
+        def cli_writer():
+            menu_active = False
+            while term.alive:
+                try:
+                    c = term.console.getkey()
+                except KeyboardInterrupt:
+                    c = '\x03'
+                if not term.alive:
+                    break
+                if menu_active:
+                    term.handle_menu_key(c)
+                    menu_active = False
+                elif c == term.menu_character:
+                    menu_active = True # next char will be for menu
+                elif c == '\x02' or  c == '\x12': # ctrl+b/ctrl+r sendbreak
+                    cdc_reset(term.serial)
+                elif c == '\x03' or c == '\x1d': # ctrl+c/ctrl+]
+                    term.stop()
+                    term.alive = False
+                    break
+                elif c == '\x05': # ctrl+e
+                    term.echo = not term.echo
+                elif c == '\x08': # ctrl+e
+                    print term.get_help_text()
+                elif c == '\t': # tab/ctrl+i
+                    term.dump_port_settings()
+                else:
+                    text = c
+                    for transformation in term.tx_transformations:
+                        text = transformation.tx(text)
+                    term.serial.write(term.tx_encoder.encode(text))
+                    if term.echo:
+                        echo_text = c
+                        for transformation in term.tx_transformations:
+                            echo_text = transformation.echo(echo_text)
+                        term.console.write(echo_text)
+        term.writer = cli_writer
+        action('--- Terminal on {p.name} - {p.baudrate},{p.bytesize},{p.parity},{p.stopbits} ---\n'.format(p=term.serial))
+        action('--- Quit: CTRL+C | Reset: CTRL+B | Echo: CTRL+E ---')
+        action('--- Info: TAB    | Help:  Ctrl+H | Menu: Ctrl+T ---')
+        term.start()
+        try:
+            term.join(True)
+        except KeyboardInterrupt:
+            pass
+        term.join()
+        term.close()
+
+    result = False
+    serial_port = get_instance(port, baudrate=baudrate, timeout=timeout)
+    if serial_port:
+        serial_port.reset_input_buffer()
+        if reset:
+            cdc_reset(serial_port)
+            result = True
+
+        if sterm:
+            if not serial_port.is_open:
+                serial_port = get_instance(port, baudrate=baudrate, timeout=timeout)
+            try:
+                cdc_term(serial_port)
+                result = True
+            except:
+                pass
+
+    return result
+
+
 # Subparser handling
 parser = argparse.ArgumentParser(prog='mbed',
     description="Command-line code management tool for ARM mbed OS - http://www.mbed.com\nversion %s\n\nUse 'mbed <command> -h|--help' for detailed help.\nOnline manual and guide available at https://github.com/ARMmbed/mbed-cli" % ver,
@@ -2366,12 +2463,13 @@ def status_(ignore=False):
     dict(name='--build', help='Build directory. Default: build/'),
     dict(name=['-c', '--clean'], action='store_true', help='Clean the build directory before compiling'),
     dict(name=['-f', '--flash'], action='store_true', help='Flash the built firmware onto a connected target.'),
+    dict(name=['-s', '--sterm'], action='store_true', help='Open serial terminal after compiling. Can be chained with --flash'),
     dict(name=['-N', '--artifact-name'], help='Name of the built program or library'),
     dict(name=['-S', '--supported'], dest='supported', const=True, choices=["matrix", "toolchains", "targets"], nargs="?", help='Shows supported matrix of targets and toolchains'),
     dict(name='--app-config', dest="app_config", help="Path of an app configuration file (Default is to look for 'mbed_app.json')"),
     help='Compile code using the mbed build tools',
     description="Compile this program using the mbed build tools.")
-def compile_(toolchain=None, target=None, profile=False, compile_library=False, compile_config=False, config_prefix=None, source=False, build=False, clean=False, flash=False, artifact_name=None, supported=False, app_config=None):
+def compile_(toolchain=None, target=None, profile=False, compile_library=False, compile_config=False, config_prefix=None, source=False, build=False, clean=False, flash=False, sterm=False, artifact_name=None, supported=False, app_config=None):
     # Gather remaining arguments
     args = remainder
     # Find the root of the program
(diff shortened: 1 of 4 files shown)


theotherjimmy:
It seems odd to me that compile would open a serial terminal. I fear we are turning `mbed compile` into `mbed do`. That being said, I can't think of a verb that means build it then run it.

mbed/mbed.py
```diff
@@ -34,6 +34,7 @@
 import zipfile
 import argparse
 import tempfile
+from mbed_cdc import mbed_cdc
```
theotherjimmy: This import should probably be scoped in some way. maybe `from mbed.mbed_cdc` or `from .mbed_cdc`. Further, this seems very redundant: `from thing import thing`.

mbed/mbed.py
```diff
@@ -2451,23 +2453,24 @@ def compile_(toolchain=None, target=None, profile=False, compile_library=False,
                   + args,
                   env=env)
 
+            if flash or sterm:
+                detected = program.detect_target()
+                try:
+                    from mbed_host_tests.host_tests_toolbox import flash_dev
+                except (IOError, ImportError, OSError):
+                    error("The '-f/--flash' option requires that the 'mbed-greentea' python module is installed.\nYou can install mbed-greentea by running 'pip install mbed-greentea'.", 1)
```
theotherjimmy: This is not strictly true. The `-f/--flash` option requires `mbed-host-tests`.

mbed/mbed_cdc.py
```diff
@@ -0,0 +1,111 @@
+
```
theotherjimmy: License?

mbed/mbed_cdc.py
```diff
@@ -0,0 +1,111 @@
...
+def mbed_cdc(port, reset=False, sterm=False, baudrate=9600, timeout= 10, print_term_header=True):
```
theotherjimmy: Could we drop `mbed` from the name of this function? it adds no meaning here. I think this is more a `run_serial_term` or something similar.

mbed/mbed_cdc.py
```diff
@@ -0,0 +1,111 @@
...
+    try:
+        from serial import Serial, SerialException
+        import serial.tools.miniterm as miniterm
+    except (IOError, ImportError, OSError):
+        return False
+
+    def get_instance(*args, **kwargs):
```
theotherjimmy: This function does not capture any state. Could you move it outside of the body of the `mbed_cdc` function?

mbed/mbed_cdc.py
```diff
@@ -0,0 +1,111 @@
...
+    def cdc_reset(serial_instance):
+        try:
+            serial_instance.sendBreak()
+        except:
+            try:
+                serial_instance.setBreak(False) # For Linux the following setBreak() is needed to release the reset signal on the target mcu.
+            except:
+                result = False
+
+    def cdc_term(serial_instance):
```
theotherjimmy: I'm pretty sure that this function does not capture any state either. Could you move this function out of `mbed_cdc`?

mbed/mbed_cdc.py
```diff
@@ -0,0 +1,111 @@
...
+                    console_print('[MENU]')
+                    menu_active = True # next char will be for menu
+                elif c == '\x02' or  c == '\x12': # ctrl+b/ctrl+r sendbreak
+                    console_print('[RESET]')
+                    cdc_reset(term.serial)
+                elif c == '\x03' or c == '\x1d': # ctrl+c/ctrl+]
+                    console_print('[QUIT]')
+                    term.stop()
+                    term.alive = False
+                    break
```
theotherjimmy: This break is not needed; the loop already checks for `term.alive`

mbed/mbed_cdc.py
```diff
@@ -0,0 +1,111 @@
...
+            serial_port.flush()
+        except Exception as e:
+            error("Unable to open serial port connection to \"%s\"" % port)
+            return False
+        return serial_port
+
+    def cdc_reset(serial_instance):
+        try:
+            serial_instance.sendBreak()
+        except:
```
theotherjimmy: This except will catch `ctrl+c`. You probably did not mean to do that.

mbed/mbed_cdc.py
```diff
@@ -0,0 +1,111 @@
...
+            try:
+                serial_instance.setBreak(False) # For Linux the following setBreak() is needed to release the reset signal on the target mcu.
+            except:
```
theotherjimmy: This except will catch `ctrl+c`. You probably did not mean to do that.

mbed/mbed_cdc.py
```diff
@@ -0,0 +1,111 @@
...
+        term.close()
+
+
+    result = False
+    serial_port = get_instance(port, baudrate=baudrate, timeout=timeout)
+    if serial_port:
+        serial_port.reset_input_buffer()
+        if reset:
+            cdc_reset(serial_port)
+            result = True
```
theotherjimmy: `cdc_reset` may set result to `False` this statement ignores that assignment, effectively throwing away the reset failure.

mbed/mbed.py
screamerbg: I'd rather encourage the user to install `mbed-greentea` which will pull in all the needed modules behind it. In this case, the product is `mbed-greentea`, not host-tests.

mbed/mbed_cdc.py
screamerbg: It does - see `port` in printed message.

mbed/mbed_cdc.py
```diff
@@ -0,0 +1,111 @@
...
+                        text = transformation.tx(text)
+                    term.serial.write(term.tx_encoder.encode(text))
+                    if term.echo:
+                        echo_text = c
+                        for transformation in term.tx_transformations:
+                            echo_text = transformation.echo(echo_text)
+                        term.console.write(echo_text)
+        term.writer = input_handler
+
+        if print_term_header:
```
theotherjimmy: If you're referring to this by `port`, why not just pass it in?

mbed/mbed.py
theotherjimmy: Sure, that works. If `mbed-host-tests` is not something the user should ever install standalone, I'll merge it with greentea.

mbed/mbed_cdc.py
theotherjimmy: port is the first parameter.

mbed/mbed_cdc.py
screamerbg: It uses other variables from the wrapper function, e.g. imports and cdc_reset()

mbed/mbed_cdc.py
screamerbg: Then perhaps you could fix it in `mbed-host-tests` as well? :) https://github.com/ARMmbed/htrun/blob/master/mbed_host_tests/host_tests_plugins/module_reset_mbed.py#L77

mbed/mbed_cdc.py
screamerbg: See previous comment

mbed/mbed_cdc.py
screamerbg: But it's not a standard serial terminal and rather for Mbed needs. So I'd rather not use a generic name. Also the whole file would rather go in https://github.com/ARMmbed/mbed-sterm, and this function name will be the contract.

mbed/mbed_cdc.py
screamerbg: Not sure what you meant. Could you elaborate?

mbed/mbed_cdc.py
theotherjimmy: Yep! Just a second.

mbed/mbed.py
theotherjimmy: This is not a must have, this works in python 2

mbed/mbed_cdc.py
theotherjimmy: Please ignore, I was mistaken.

mbed/mbed.py
theotherjimmy: In case this was not clean: this is resolved.

mbed/mbed_cdc.py
theotherjimmy: Understood. This is resolved.

mbed/mbed_cdc.py
theotherjimmy: This is just a minor style comment. Not a blocker.

theotherjimmy:
@screamerbg All that remains to be fixed is the two `except:`s

mbed/mbed_cdc.py
screamerbg: Added. Thanks

SenRamakri:
Why is this only an option to detect/compile? Can we start it by itself? like, "mbed startserial <other options if multiple targets are detected>" OR "mbed sterm".
Also from a developer standpoint, people tend to keep separate terminal to run the builds. So it would be useful if I can do "mbed sterm" separately. And yes, as @sg- mentioned we can potentially expand on this to be used with Binary/ID-based tracing so that it automatically pick the right elf file to do the ID-string mapping.

theotherjimmy:
@SenRamakri How is `mbed sterm` different from `mbed detect -s`? both have to detect attached targets before attaching the terminal.

SenRamakri:
If "mbed detect -s" take options in case of multiple targets, then yes, we don't need "mbed sterm".

theotherjimmy:
@SenRamakri What do you mean by "take options in case of multiple targets"?  What options would it take? how do they apply to multiple targets?

SenRamakri:
Offline discussion with @theotherjimmy here confirmed that "mbed detect" does pick one of many targets if required. So, "mbed detect -s" does support what I was looking for. Thanks for confirming, @theotherjimmy

yennster:
So glad this is taking off @screamerbg ! I'll take a look at the README tomorrow 😄

yennster:
@screamerbg Is there a way to change the baud rate for the serial monitor? Right now it looks like it defaults to [`9600`](https://github.com/screamerbg/neo/blob/e9d77ea584b28fcf679fa06fc24943bb0245013e/mbed/mbed_cdc.py#L22).

jupe:
Does this feature support terminal codes (ANSI/VT100) ?

screamerbg:
@yennster You can change the baudrate form within the terminal using Ctrl+T + B combo. There's no commandline flag (yet)

screamerbg:
@jupe The implementation invokes PySerial miniterm. This is from miniterm's documentation:

_This is a console application that provides a small terminal application._

_Miniterm itself does not implement any terminal features such as VT102 compatibility. However it may inherit these features from the terminal it is run. For example on GNU/Linux running from an xterm it will support the escape sequences of the xterm. On Windows the typical console window is dumb and does not support any escapes. When ANSI.sys is loaded it supports some escapes._

http://pyserial.readthedocs.io/en/latest/tools.html#module-serial.tools.miniterm

On Mac terminal it provides a fully featured ANSI support. Had a chance to try it with an ANSI  embedded program @stevew817 wrote.

jupe:
I tried this with [mbed-client-cli](https://github.com/armmbed/mbed-client-cli) which provides ANSI support. Couple comments / proposals:
* could echo be disabled by default ? 
* could there be cli parameter for baudrate ?
* When serial port dies it raises `SerialException` - could it be handled more carefully ?
* If I press backspace help is printed unexpectedly -> cannot remove characters from cli interface
* when calling `mbed detect -s` it prints following logs:
    ```
    [mbed] WARNING: The mbed OS tools were not found in "/tmp". 
    [mbed] WARNING: Limited information will be shown about connected mbed targets/boards
    ```
    What this `Limited information` means in practice ?
* `CTRL+T` `CTRL+B` causes `IOError: [Errno 25] Inappropriate ioctl for device`  and application stuck
   This combination is not mentioned in help, but it's easy to do this mistake because ``CTRL+T` `b` is used to change baudrate.

screamerbg:
Follow up PR #664

PR #652, opened 2018-04-03
Added README info for Serial Terminal

yennster:
This PR adds help information to the README file for the new Serial Terminal feature (https://github.com/ARMmbed/mbed-cli/pull/650)

It includes:
- 2 usage examples for `mbed compile` and `mbed detect`
- List of available keyboard shortcuts

@screamerbg

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,11 +183,11 @@ $ cd mbed-os
 $ mbed releases
 mbed-os (#182bbd51bc8d, tags: latest, mbed-os-5.6.5)
   ...
-  * mbed-os-5.6.0 
-  * mbed-os-5.6.1 
-  * mbed-os-5.6.2 
-  * mbed-os-5.6.3 
-  * mbed-os-5.6.4 
+  * mbed-os-5.6.0
+  * mbed-os-5.6.1
+  * mbed-os-5.6.2
+  * mbed-os-5.6.3
+  * mbed-os-5.6.4
   * mbed-os-5.6.5  <- current
 ```
 
@@ -408,6 +408,7 @@ The arguments for *compile* are:
 * `--config` to inspect the runtime compile configuration (see below).
 * `-S` or `--supported` shows a matrix of the supported targets and toolchains.
 * `-f` or `--flash` to flash/program a connected target after successful compile.
+* `-s` or `--sterm` to open the connected target's serial terminal after successful compile. This option can be chained with `-f` or `-r` so the new firmware image will be flashed to the target or reset, then open the serial terminal.
 * `-c ` to build from scratch, a clean build or rebuild.
 * `-j <jobs>` to control the compile processes on your machine. The default value is 0, which infers the number of processes from the number of cores on your machine. You can use `-j 1` to trigger a sequential compile of source code.
 * `-v` or `--verbose` for verbose diagnostic output.
@@ -479,10 +480,10 @@ $ mbed compile --config -t GCC_ARM -m K64F --prefix target --prefix app
 
 #### Macros
 
-You can specify macros in your command-line using the -D option. For example:
+You can specify macros in your command-line using the `-D` option. For example:
 
 ```
-$ mbed compile -t GCC_ARM -m K64F -c -DUVISOR_PRESENT
+$ mbed compile -t GCC_ARM -m K64F -c -D UVISOR_PRESENT
 ```
 
 #### Compiling in debug mode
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -408,7 +408,8 @@ The arguments for *compile* are:
 * `--config` to inspect the runtime compile configuration (see below).
 * `-S` or `--supported` shows a matrix of the supported targets and toolchains.
 * `-f` or `--flash` to flash/program a connected target after successful compile.
-* `-c ` to build from scratch, a clean build or rebuild.
+* `-s` or `--sterm` to open the connected target's serial terminal. This option can be chained with `-f` or `-r` so the new firmware image will be flashed to the target or reset, then open the serial terminal.
```
sg-: -s and -S seem confusing. I get -fr for flash and reset but seems that ending in a terminal should be an explicit --sterm or better --terminal or --term

theotherjimmy: We would have to change the code than handles arguments to make that change. That's out of scope for this PR, probably.

theotherjimmy: I personally don't like using the word "terminal" here. It's a tad off, as it's specifically a hardware thing https://en.wikipedia.org/wiki/Computer_terminal . What were talking about is really a terminal emulator. We could also claim that daplink is the terminal, and that we are connecting to it, so `--connect` might also make sense.

yennster: @theotherjimmy In terms of other tutorials (like the [offline quickstart](https://os.mbed.com/docs/v5.8/tutorials/quick-start-offline.html#printf)) we refer to viewing serial port output as opening up a "serial terminal". I personally think calling it an emulator is confusing. Thoughts? Maybe "serial port viewer"?

theotherjimmy: In terms of how the word "terminal" is used these days, it probably means "something that takes input, accepts output and looks like a command line". That's why I mentioned that it was a personal pref. Also, calling it an "emulator" is confusing because you omit "terminal" and the user is left without the context of what is being emulated.

-------------

I think that "serial port viewer" may have the same problem that "sterm" and "serial terminal" has: it starts with an "s", making the short option `-s`, which conflicts with `-S`. Further, I would rather we use terminal here, as it's a better understood word for this than "port viewer". Especially as "port viewer" might imply to a user that they cannot interact with the device.

screamerbg: Agreed with @yennster. Serial terminal is rather established term (check mini term).

Terminal emulator is very confusing - we're no longer in the 80's/90s when a terminal meant a monitor and a keyboard attached to a system, usually part of a bigger system.

"connect" doesn't make sense either, especially in the light of terminology in the IoT world.

screamerbg: https://en.wikipedia.org/wiki/Terminal_mode

theotherjimmy: @screamerbg I claimed that it was a personal pref.

-----

That Terminal mode article links to "Terminal Emulator" and makes no assertion that a Terminal is not a hardware device, or that a Terminal Emulator would not be capable of emulating the same behavior.

-----

I think that "serial terminal" is inconsistent with `compile`, `detect`, `--flash` and  `--reset`, as it's a noun instead of a verb. I proposed "connect" because it fit the pattern of verbs.

screamerbg: @sg- Agreed that `-s` and `-S` are somewhat confusing. We can speculate which one would be more used `-S` (for supported) or `-s` for serial terminal, yet for backwards compatibility we can't deprecate `-S`

Perhaps we could rethink the workflow itself. For example compile+flash+term can be chained via `mbed compile -f --term` (which is fine with me), but for flash+terminal only a user could use `mbed term -f` instead of `mbed detect -f -s`. E.g. introduce verb `mbed term` with all required options, like baudrate, encoding, etc. instead of cramming it in `mbed detect`.

What do you think?

theotherjimmy: I'm on board with adding another verb for this extended workflow. We might have contention about naming that verb though. `term` seems like a noun to me.

screamerbg: `mbed terminal`?

Note that `mbed status` is not a verb, nor is `mbed releases`

theotherjimmy: Well you have me there :D The `-s` vs `-S` conflict exists on `mbed compile` only, as far as I know. Are you suggesting that `mbed term`/`mbed terminal` builds the code too? Maybe through a switch? Maybe as opt-out?

yennster: `mbed monitor`? 😆

kjbracey-arm: Is "console" the word you're looking for?

yennster: `mbed console` sounds good, but then the option `-c` would conflict with the exisiting `-c` for a clean build

yennster:
See revised PR https://github.com/ARMmbed/mbed-cli/pull/664

screamerbg:
@yennster The Serial Terminal functionality made it in the Mbed CLI 1.6 release. Do you think that this PR is still needed?

PR #666, opened 2018-05-08
Add support for --ignore argument in test.py

andrewleech:
Simply passes the parameter through when building with test.py

Supports and requires https://github.com/ARMmbed/mbed-os/pull/6833

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -2489,9 +2489,10 @@ def compile_(toolchain=None, target=None, profile=False, compile_library=False, 
     dict(name='--test-spec', dest="test_spec", help="Path used for the test spec file used when building and running tests (the default path is the build directory)"),
     dict(name='--app-config', dest="app_config", help="Path of an app configuration file (Default is to look for 'mbed_app.json')"),
     dict(name='--test-config', dest="test_config", help="Path or mbed OS keyword of a test configuration file. Example: ethernet, odin_wifi, or path/to/config.json"),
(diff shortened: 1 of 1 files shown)


screamerbg:
@andrewleech Thanks for this. Looking good to me.

@theotherjimmy @cmonr Any concerns?

theotherjimmy:
@screamerbg @andrewleech We should have the same feature in compile and export, for consistency.

screamerbg:
Would like to poke @sg- for his take on this.

screamerbg:
@andrewleech Note that this won't work prior to Mbed OS 5.9 (after https://github.com/ARMmbed/mbed-os/pull/6833 patch is merged), therefore the option help should explicitly mention this.

sg-:
Is there a use case for this other than ignoring main when running tests? This certainly is a simple solution but would worry about the annoyance of not remembering what file contains the main function.

@bridadan Do you recall the previous thinking around solutions to multiple main functions?

andrewleech:
I've also used it since to build tests with rtos (--exclude main) and then compile app without rtos (--exclude mbed-os/rtos)

bridadan:
> @bridadan Do you recall the previous thinking around solutions to multiple main functions?

I believe the option we suggested at the time was to provide a different entry point for tests. Something like `test_main` or something instead of `main`. This could be enabled by specifying an additional `define` when compiling with `mbed test`.

One thing to keep in mind with that approach though is that any global static defines in your app's `main` file would still be brought in for each test (you would pay a RAM/ROM penalty). Short of knowing where that file is and excluding it from the build I don't think there's a way we can prevent that.

PR #668, opened 2018-05-11
Update CircleCI to run parallel Python jobs

cmonr:
Replaces circle.yml with new CircleCI format.

py3 test is disabled until https://github.com/ARMmbed/mbed-cli/pull/667 comes in.

Updated config format to version 2.


.circleci/config.yml
```diff
@@ -0,0 +1,183 @@
+version: 2
+jobs:
+  py2:
+    machine: true
+    steps:
+      - run: sudo add-apt-repository -y ppa:terry.guo/gcc-arm-embedded
+      - run: sudo apt-get -y update
+      - run: sudo apt-get -y install gcc-arm-none-eabi=4.9.3.2015q3-1trusty1
```
bridadan: Can we update this to GCC 6.x? Pretty sure Mbed OS doesn't [recommend using 4.9.x anymore](https://os.mbed.com/docs/v5.8/tools/index.html).

.circleci/config.yml
```diff
@@ -0,0 +1,183 @@
...
+          mbed target -S
+          mbed export -S
+      - run: |-
+          cd .tests/supported-tests
+          echo "Testing mbed-os-5.1.0"
+          mbed update --clean "mbed-os-5.1.0"
+          mbed compile -S
+          mbed toolchain -S
+          mbed target -S
+          mbed export -S
```
bridadan: Instead of copy and pasting these commands, can the common ones be extracted to either a script or circleci job?

cmonr: Indeed. I also think there should be one test per feature release of mbed-os, but that was out of scope for this PR.

.circleci/config.yml
cmonr: Woah. You're right, but I'd prefer to change only what's needed for the topic at hand.

Definitely going to need to submit another PR to get that updated.

.circleci/config.yml
```diff
@@ -103,81 +106,41 @@ jobs:
...
       - run: py.test test
       - run: mbed toolchain -G GCC_ARM
       - run: mbed target -G K64F
       - run: mbed config -G protocol ssh
+
       - run: mkdir .tests
       - run: cd .tests && mbed new new-test
+      - run: |-
+          cd .tests/new-test/mbed-os
+          git fetch origin refs/pull/6886/head:pr6886
```
cmonr: Fyi: There was one last issue in mbed-os that this testing revealed, and is fixed with https://github.com/ARMmbed/mbed-os/pull/6886.

Once that PR is in, these two lines will need to be updated to origin/master until 5.9.0 is released to make sure the updates are pulled in correctly.

cmonr:
Forgot that this PR doesn't have the mbed-cli py3 fixes.

This is how far the CircleCI script gets with py3 before failing, assuming this PR and https://github.com/ARMmbed/mbed-cli/pull/667 are merged together: https://circleci.com/gh/cmonr/mbed-cli/157


The problem appears to be with https://os.mbed.com/teams/Morpheus/code/mbed-os/ when running the hg tests. Afaik, this won't be fixed until an mbed-os release is made.

I could use some help in seeing if this can be worked around, similar to the line 118 workaround for git.

.circleci/config.yml
screamerbg: Updating GCC 6.x would only work for newer Mbed OS, but older cases will fail (like Morpheus). And the Morpheus routine is used for comprehensive multi-level dependencies handling, including removal of libraries etc.

bridadan: That make sense, but we should be testing with the recommended toolchain for each Mbed OS release. So if that means installing two different versions of GCC then we may need to do that.

setup.py
```diff
@@ -32,5 +32,5 @@
             'mbed-cli=mbed.mbed:main',
         ]
     },
-    python_requires='>=2.7.10, <3',
+    python_requires='>=2.7.10',
```
screamerbg: Perhaps this change should come with the other PR.

.circleci/config.yml
```diff
@@ -0,0 +1,145 @@
+version: 2
```
screamerbg: Could you please update the `circle_tests.py` script which allows to run all commands from the circle.yml file and emulate circle CI test run?

setup.py
cmonr: Agreed. Will revert this.

cmonr:
@bridadan @screamerbg @theotherjimmy Mind rereviewing?

.circleci/config.yml
cmonr: Now works across both versions!

screamerbg:
Looks great!

PR #674, opened 2018-05-24
sterm fixes for py3

cmonr:
Tested with py3 using pipenv virtual environment. Same code works in both py2 and py3.


screamerbg:
@cmonr This is great! Could you please remove the print line? (was mistake on my side)

cmonr:
Done!

Issue #677, opened 2018-05-25
v1.7.0 is broken on Python 2


amq:
When `v1.7.0` got released today, my CI builds started to fail with:

```
Traceback (most recent call last):
  File "/usr/local/bin/mbed", line 7, in <module>
    from mbed.mbed import main
  File "/usr/local/lib/python2.7/dist-packages/mbed/mbed.py", line 22, in <module>
    from future.builtins.iterators import zip
ImportError: No module named future.builtins.iterators
```

`Dockerfile` to replicate the issue:

```
FROM debian:stretch-slim
RUN apt update && \
    apt install -y lbzip2 ca-certificates wget python-pip git mercurial --no-install-suggests && \
    rm -rf /var/lib/apt/lists/*
RUN pip install mbed-cli
RUN mbed config -G GCC_ARM_PATH "/usr/bin"
```

amq:
Strangely, Python 3 also doesn't seem to work:

```
Traceback (most recent call last):
  File "/usr/local/bin/mbed", line 7, in <module>
    from mbed.mbed import main
  File "/usr/local/lib/python3.5/dist-packages/mbed/mbed.py", line 22, in <module>
    from future.builtins.iterators import zip
ImportError: No module named 'future'
```

```
FROM debian:stretch-slim
RUN apt update && \
    apt install -y lbzip2 ca-certificates wget python3-pip git mercurial --no-install-suggests && \
    rm -rf /var/lib/apt/lists/*
RUN pip3 install mbed-cli
RUN mbed config -G GCC_ARM_PATH "/usr/bin"
```

theotherjimmy:
@amq It looks like we're depending on the `future` package in pip, but it was not added to the dependencies in setup.py.

theotherjimmy:
As a workaround you should be able to pip install future.

amq:
Installing future fixes the problem.

amq:
There is a different problem when building:

```
Traceback (most recent call last):
  File "/home/amq/project/mbed-os/tools/make.py", line 41, in <module>
    from tools.tests import TESTS, Test, TEST_MAP
  File "/home/amq/project/mbed-os/tools/tests.py", line 18, in <module>
    from tools.data.support import DEFAULT_SUPPORT, CORTEX_ARM_SUPPORT
  File "/home/amq/project/mbed-os/tools/data/support.py", line 17, in <module>
    from tools.targets import TARGETS
[...]

PR #678, opened 2018-05-26
Fix future python module dependency

screamerbg:
And bump to 1.7.1 release


screamerbg:
Without this Mbed CLI cannot be run on new systems (#677)

PR #679, opened 2018-05-26
Fix build directories name casing

screamerbg:
This is a small fix to ensure that `mbed compile` and `mbed test` output directories will use upper casing for case-sensitive operating systems. This avoids the problem where `mbed compile -m k64f` and `mbed compile -m K64F` produces 2 separate output build folders.

Related to #529 and #530.

cc @JammuKekkonen @korjaa 
Also cc @OPpuolitaival @studavekar regarding possible CI related issues/questions.


theotherjimmy:
@screamerbg Working twards 1.8.0 I see!

bridadan:
Nice! 😄

Issue #436, opened 2017-02-15
Creating a new program for mbed OS 2, the mbed library isn't downloaded


eavelardev:
macOS Sierra
mbed 1.0.0

When I create a new mbed OS 2 program the mbed library isn't downloaded automatically. I execute `mbed update` inside the project folder to resolve.

screamerbg:
@eavelardev What command do you use?

eavelardev:
I use `mbed new my_project_name --mbedlib`

screamerbg:
Hi,

I just used `mbed new <program> --mbedlib` and worked fine. Here is the output


```
D:\1>mbed --version
1.0.0

D:\1>mbed new test-classic --mbedlib
[mbed] Creating new program "test-classic" (git)
[mbed] Adding library "mbed" from "https://mbed.org/users/mbed_official/code/mbed/builds" at latest revision in the current branch
[mbed] Downloading library build "ef9c61f8c49f" (might take a minute)
[mbed] Unpacking library build "ef9c61f8c49f" in "D:\1\test-classic\mbed"
[mbed] Updating reference "mbed" -> "https://mbed.org/users/mbed_official/code/mbed/builds/ef9c61f8c49f"
[mbed] Couldn't find build tools in your program. Downloading the mbed 2.0 SDK tools...
```

screamerbg:
@eavelardev bump

screamerbg:
Closing as we're not able to reproduce. Please reopen if needed.

michaelVisimid:
The issue still exists with version 1.4. This is pretty easily reproduced and still an issue. After you changed to your test program's directory you should have either run tree to see that the mbed libraries are not downloaded, or added a main.cpp which would have showed you the exact same thing. I'm not quite sure how the recreate steps taken would have recreated the issue as at no point were the libraries actually checked for. Below is the output from the command prompt where I should that the issue still exists and that adding main.cpp and importing mbed-dev allows the program to compile. For that matter, why doesn't the CLI create a bare minimum main.cpp that looks something like this:

`#include "mbed.h"

int main() {
    while(1) {      
    }
}

`
Here is the command prompt output which is easily recreated with the use of either tree or attempting to compile with main.cpp. 

`C:\Users\mbaker\Desktop\test>py -2.7 -m mbed --version
1.4.0

C:\Users\mbaker\Desktop\test>py -2.7 -m mbed new test-classic --mbedlib
[mbed] Creating new program "test-classic" (git)
[...]

PR #680, opened 2018-05-26
Fix mbed2 builds download when creating new mbed2 program

screamerbg:
This is a fix for reported issue #436

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -353,7 +353,10 @@ class Bld(object):
         try:
             Bld.init(path)
             with cd(path):
-                Bld.seturl(url+'/tip')
+                rev = Hg.remoteid(m.group(1), 'tip')
+                if not rev:
+                    error("Unable to fetch library build information")
+                Bld.seturl(url+'/'+rev)
         except Exception as e:
             if os.path.isdir(path):
                 rmtree_readonly(path)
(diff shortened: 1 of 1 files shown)


mbed/mbed.py
```diff
@@ -1236,6 +1236,9 @@ def clone(self, url, path, rev=None, depth=None, protocol=None, offline=False, *
                     continue
                 try:
                     scm.clone(url, path, depth=depth, protocol=protocol, **kwargs)
+                    if scm.name == 'bld':
+                        with cd(path):
+                            scm.update()
```
theotherjimmy: Does this need to happen only for `bld` or is this also valid for other scms?

screamerbg: It's a bld specific thing. 

With real SCMs (Git/Hg) you want to get as much as possible from the index from the clone() operation as you can then freely change between tags. Bld is emulating SCM though, we don't want anything checked out during Bld.clone() with no revision, because this will deliver a 300MB codebase. Bld.clone() rather "initializes" a build SCM, and then needs an update() to either latest or a specific revision.

theotherjimmy: Alright. I would like to revisit this, as it seems like we're breaking the SCM abstraction here.

screamerbg: @theotherjimmy I actually rewrote so we're not breaking the SCM abstraction. Check what was merged in.

theotherjimmy: Yeah, I noticed. That's why I merged!

PR #689, opened 2018-06-04
Added warning message about Mbed OS Py3 compatability

cmonr:
New output:
```
$ mbed new .
[mbed] WARNING: Usage of Python 3 requires Mbed OS >= v5.9
---
[mbed] Creating new program "test" (git)
[mbed] Adding library "mbed-os" from "ssh://git@github.com/ARMmbed/mbed-os" at branch/tag "latest"
...
```

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -2989,7 +2989,7 @@ def main():
 
     # Python 3 backwards compatability warning
     if sys.version_info[0] == 3:
-        warning("Usage of Python 3 requires Mbed OS >= v5.9")
+        warning("If you're using Python 3 with Mbed OS 5.8 and earlier versions, Python errors will occur when compiling, testing and exporting")
 
     pargs, remainder = parser.parse_known_args()
     status = 1
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -2987,6 +2987,10 @@ def main():
         log(ver+"\n")
         sys.exit(0)
 
+    # Python 3 backwards compatability warning
+    if sys.version_info[0] == 3:
+        warning("Usage of Python 3 requires Mbed OS >= v5.9")
+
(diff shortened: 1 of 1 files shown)


MarceloSalazar:
The warning is great.
But can we still catch the traceback error?

cmonr:
@MarceloSalazar The particular traceback error that you hit will actually be resolved once https://github.com/ARMmbed/mbed-os/pull/7092 is merged. 

Continuing the issue conversation in here,

> This should be ok as quick/initial fix. My main interest is to catch the traceback error and display a message.

In what regard? Tracebacks can occur regardless of Python version.

MarceloSalazar:
> The particular traceback error that you hit will actually be resolved once ARMmbed/mbed-os#7092 is merged.

I understand 7092 will fix the problem from 5.9+, which is good.

But I believe we'll still hit the error with <=5.8, unless we include a fix in CLI.
Is there anything we could do in CLI to catch the trace back for this particular use case?

MarceloSalazar:
Btw, I'm happy to get this PR done and continue the conversation in 686.

screamerbg:
@cmonr Could you expand the message to
```
WARNING: You're using Python 3 which requires Mbed OS 5.9 and above. If you're using Mbed OS 5.8 and earlier versions, you will get Python errors when compiling, testing and exporting.
```

theotherjimmy:
@cmonr Could we print this message only on an error? that way we don't clutter up _perfectly working projects_

theotherjimmy:
If it's harder, feel free to clutter things  up for now.

screamerbg:
@theotherjimmy Let's get this in as a short-term solution. I'd like to add a mini-patch that detects the version by reusing the functionality from `mbed releases` and triggers a message if incompatibility is detected.

screamerbg:
LGTM! Thanks

PR #699, opened 2018-06-12
Use relative paths for source and build dirs when testing

bridadan:
This was discovered when doing some investigation for https://github.com/ARMmbed/mbed-os/issues/6861.

Currently, `mbed test` is passing absolute paths for the `--source` and `--build` options. This path is not shortened by the Mbed OS tools when compiling, so the include paths are much longer than they need to be. This alters `mbed test` to instead pass relative paths to both `--source` and `--build`.

While this won't fix the issue completely, it will drastically reduce the include path lengths when using `mbed test`.

*Steps to reproduce**

Before this patch, if you run `mbed test -m NRF52_DK -t GCC_ARM` in the root of `mbed-os` at commit `ed9a1f1327f5208fd462c5a6938696ee15e393f1` (recent commit off master branch), it will immediately fail to compile.

After applying this patch, you will be able to compile and test.

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -2589,12 +2589,12 @@ def test_(toolchain=None, target=None, compile_list=False, run_list=False, compi
     with cd(program.path):
         # Setup the source path if not specified
         if not source or len(source) == 0:
-            source = [program.path]
+            source = [os.path.relpath(program.path, orig_path)]
 
         # Setup the build path if not specified
         build_path = build
         if not build_path:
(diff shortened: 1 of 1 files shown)


screamerbg:
I like the patch but I think there's bigger fundamental problem with long paths on Windows and the Mbed OS codebase paths, etc. See email thread from @dlfryar. Not sure whether this helps or smears the problem or postpones it for another time.

bridadan:
This change brings the way paths are handled in line with the `compile` command. So more of a clarity and consistency thing than anything else.

Issue #714, opened 2018-07-04
Program wrong platform


MarceloSalazar:
Issue: defect.

I compile an application and want to program a platform using this option:
```mbed compile -t <toolchain> -m <platform> -f```

However, if have a **different** platform connected, that's ignored and the binary I've just compiled is used to program the wrong platform.

Solution / proposal:

It should detect that I don't have that platform detected and display the corresponding error.

ciarmcom:
ARM Internal Ref: MBOTRIAGE-1062

PR #717, opened 2018-07-11
Fix: Firmware flashing routine

### LINKED ISSUE #713: Program one of multiple platforms connected
This is an enhancement request.

I have many platforms connected to the USB port.

I'd like to compile and program one platform using this command:
```mbed compile -t <toolchain> -m <platform> -f```
This is very useful because finds the binary, detects the platform and programs it.

However, there is currently a limitation:
```[mbed] ERROR: Only 1 target board should be connected to your system.```

screamerbg:
Fix firmware flashing routine to flash the correct board/target connected to the user system.

Fixes #713 and #714

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1583,7 +1583,7 @@ class Program(object):
         target = target if target else target_cfg
 
         if target and (target.lower() == 'detect' or target.lower() == 'auto'):
-            detected = self.detect_target()
+            detected = self.detect_single_target()
             if detected:
                 target = detected['name']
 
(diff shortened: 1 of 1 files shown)


cmonr:
Does this actually fix https://github.com/ARMmbed/mbed-cli/issues/713? I would expect that fix would end up removing line 1630.

screamerbg:
@cmonr Perhaps re-read #713 and try it yourself :)

theotherjimmy:
@screamerbg My reading of the code seems to indicate that it will flash and sterm to all matching connected boards in sequence. Is this the case?

theotherjimmy:
@screamerbg @cmonr  The specific error message should only happen when using the `-t auto` or `-t detect` flags.

PR #718, opened 2018-07-11
Fix: Python 3 tests not to use Mbed OS master

screamerbg:
This fixes Python 3 tests not to use Mbed OS master, but fixed version (Mbed OS 5.9.0)


screamerbg:
Temporary fix to unlock Mbed-CLI master until Mbed OS tools are fixed.
cc @theotherjimmy @cmonr

PR #719, opened 2018-07-13
Bump version to 1.7.5 and add pypi.org markdown support

screamerbg:
Add the title states


theotherjimmy:
@screamerbg Merge and release at will.

PR #723, opened 2018-07-25
Add instructions on how to compile multiple projects with one copy of Mbed OS

yennster:
This PR adds instructions to the README on how a user can use the `MBED_OS_DIR` config within Mbed CLI to manage multiple Mbed projects that use the same version of Mbed OS (and therefore the same Mbed OS directory/copy).

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,17 +222,20 @@ You can create multiple Mbed projects and use the same Mbed OS library directory
 
 ```
 $ cd <projects directory>
-$ git clone https://github.com/ARMmbed/mbed-os
+$ mbed import mbed-os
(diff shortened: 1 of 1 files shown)


yennster:
@AnotherButler If you have a spare minute can you help me with a grammar edit for this? 😎

AnotherButler:
@yennster Sure thing. Could we also add this to the os.mbed.com/docs page? https://github.com/ARMmbed/mbed-os-5-docs/blob/development/docs/tools/offline/cli-create.md

yennster:
@AnotherButler Sure! I'll put in another PR in the docs repo

yennster:
@theotherjimmy Thoughts?

PR #734, opened 2018-08-20
Support for unit testing

lorjala:
Add support for Mbed OS unit testing.

related mbed-os PR: https://github.com/ARMmbed/mbed-os/pull/7819

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -3011,6 +3011,51 @@ def help_():
     return parser.print_help()
 
 
+@subcommand('unittest',
+    dict(name='--skip-build', action='store_true', help='skip build step'),
+    dict(name='--skip-run', action='store_true', help='skip run step'),
+    dict(name='--clean', action='store_true', help='clean build data'),
+    dict(name=['-d', '--debug'], action='store_true', help='enable debug build'),
+    dict(name='--coverage', choices=['html', 'xml', 'both'], help='generate code coverage report'),
+    dict(name=['-m', '--make-program'], choices=['gmake', 'make', 'mingw32-make', 'ninja'], help='which make program to use'),
+    dict(name=['-g', '--generator'], choices=['Unix Makefiles', 'MinGW Makefiles', 'Ninja'], help='which CMake generator to use'),
+    dict(name=['-r', '--regex'], help='run tests matching regular expression'),
+    dict(name='--build-path', help='specify build path'),
+    dict(name='--new', help='generate files for a new unit test', metavar="FILEPATH"),
+    help='Create, build and run unit tests')
+def unittest_(skip_build=False, skip_run=False, clean=False, debug=False, coverage=None, make_program=None, generator=None, regex=None, build_path=None, new=None):
+    program = Program(getcwd(), False)
+    program.check_requirements(True)
+
+    mbed_os_dir = program.get_os_dir()
+    if mbed_os_dir is None:
+        error("No Mbed OS directory found.")
+    unittests_dir = os.path.join(mbed_os_dir, "UNITTESTS")
+
+    tool = os.path.join(unittests_dir, "mbed_unittest.py")
+
+    # Prepare environment variables
+    env = program.get_env()
+
+    if os.path.exists(tool):
+        popen([python_cmd, tool]
+                + (["--skip-build"] if skip_build else [])
+                + (["--skip-run"] if skip_run else [])
+                + (["--clean"] if clean else [])
+                + (["--debug"] if debug else [])
+                + (["--coverage", coverage] if coverage else [])
+                + (["--make-program", make_program] if make_program else [])
+                + (["--generator", generator] if generator else [])
+                + (["--regex", regex] if regex else [])
+                + (["--build-path", build_path] if build_path else [])
+                + (["--new", new] if new else [])
+                + (["--verbose"] if verbose else [])
+                + remainder,
+                env=env)
+    else:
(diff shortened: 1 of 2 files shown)


mbed/mbed.py
```diff
@@ -3011,6 +3011,51 @@ def help_():
...
+
+    tool = os.path.join(unittests_dir, "mbed_unittest.py")
+
+    # Prepare environment variables
+    env = program.get_env()
+
+    if os.path.exists(tool):
+        popen([python_cmd, tool]
+                + (["--skip-build"] if skip_build else [])
+                + (["--skip-run"] if skip_run else [])
```
bridadan: Instead of using `--skip-build` and `--skip-run`, could you use `--compile` and `--run` like the existing `mbed test` command? So it's an "opt-in" instead of "opt-out". It would make the testing experience more parallel with what already exists.

mbed/mbed.py
```diff
@@ -3011,6 +3011,51 @@ def help_():
     return parser.print_help()
 
 
+@subcommand('unittest',
+    dict(name='--skip-build', action='store_true', help='skip build step'),
+    dict(name='--skip-run', action='store_true', help='skip run step'),
+    dict(name='--clean', action='store_true', help='clean build data'),
```
bridadan: Can you also add the short version `-c`? This is present in the other commands that build code.

mbed/mbed.py
```diff
@@ -3011,6 +3011,51 @@ def help_():
...
+    dict(name=['-d', '--debug'], action='store_true', help='enable debug build'),
+    dict(name='--coverage', choices=['html', 'xml', 'both'], help='generate code coverage report'),
+    dict(name=['-m', '--make-program'], choices=['gmake', 'make', 'mingw32-make', 'ninja'], help='which make program to use'),
+    dict(name=['-g', '--generator'], choices=['Unix Makefiles', 'MinGW Makefiles', 'Ninja'], help='which CMake generator to use'),
+    dict(name=['-r', '--regex'], help='run tests matching regular expression'),
+    dict(name='--build-path', help='specify build path'),
```
bridadan: Can you please rename this to just `--build` to be inline with the other `mbed` commands?

mbed/mbed.py
```diff
@@ -3011,6 +3011,51 @@ def help_():
...
+@subcommand('unittest',
+    dict(name='--compile', action='store_true', dest="compile_only", help='Only compile unit tests'),
+    dict(name='--run', action='store_true', dest="run_only", help='Only run unit tests'),
+    dict(name=['-c', '--clean'], action='store_true', help='Clean the build directory'),
+    dict(name=['-d', '--debug'], action='store_true', help='Enable debug build'),
+    dict(name='--coverage', choices=['html', 'xml', 'both'], help='Generate code coverage report'),
+    dict(name=['-m', '--make-program'], choices=['gmake', 'make', 'mingw32-make', 'ninja'], help='Which make program to use'),
+    dict(name=['-g', '--generator'], choices=['Unix Makefiles', 'MinGW Makefiles', 'Ninja'], help='Which CMake generator to use'),
+    dict(name=['-r', '--regex'], help='Run tests matching regular expression'),
+    dict(name='--build', help='Build directory. Default: mbed-os/UNITTESTS/build/'),
```
bridadan: This default build directory is outside of the normal root `BUILD` directory, is this something that has already been agreed upon? @theotherjimmy

lorjala: Since unit tests use completely different build environment, they will not be part of mbed-os application and the tools used are separate from other mbed-os tools it would be better to keep the build folders separate in my opinion.

theotherjimmy: @lorjala I'm concerned that your build directory will be scanned by the build tools. How are you avoiding that?

lorjala: @theotherjimmy UNITTESTS directory has .mbedignore file, but it could be a possibility to add .mbedignore into the build directory as well during the folder creation if that is the concern.

lorjala: 😉 Defaults to `BUILD/unittests` for now. Let's see how it works.

AnotherButler:
@theotherjimmy @bridadan Could one of you please merge this?

lorjala:
@screamerbg https://github.com/ARMmbed/mbed-os/pull/7819#discussion_r211318265

AnotherButler:
@mihail Is this OK to merge?

theotherjimmy:
@screamerbg 

from @OPpuolitaival's linked comment

>we need separated command because:
"mbed test" - testing code which runs on embedded device and therefore need a target and toolchain
"mbed unittest" - testing code units in host computer using documented compilers

I think that's a satisfactory explanation. Perhaps you would like to use a different subcommand?

screamerbg:
@theotherjimmy Yeah. I'm not happy that unittesting is very internal to Mbed OS development as a product, which won't apply for 99.9% of the Mbed OS users.

theotherjimmy:
@lorjala @OPpuolitaival We have `mbed test` for everything testing. Greentea tests and Icetea tests are very different. Greentea and Unittests are also very different. Why do unittests get a seperate subcommand and icetea did not? Why is unittesting logically not testing? I think `mbed test --unit` or `mbed test --unittest` would be better here.

theotherjimmy:
I'm also open to `mbed test --host-only` or similar.

screamerbg:
LGTM. Great work @theotherjimmy!

screamerbg:
@theotherjimmy Merge at will

PR #741, opened 2018-09-03
Release 1.8.0

screamerbg:
Includes fixes and device management support.


theotherjimmy:
@screamerbg Can we wait for Icetea and unit tests? That's the Mbed OS 5.10 feature set.

Issue #660, opened 2018-04-24
Issue when several processes use mbed cache in parallel 


adamelhakham:
Hi,
We have a Jenkins job (on Ubuntu 14.04 machine) that runs `mbed deploy` multiple times in parallel in different directories. In order to speed things up we would like to use the cache feature.
However, occasionally, some of the processes run into the following issue:
```
[...]

PR #752, opened 2018-09-11
Fix: Caching thread safety

screamerbg:
This aims to fix caching thread safety in Mbed CLI. It uses directory creation which is atomic operation on all operating systems. I've tested with the script @adamelhakham wrote. Additionally @adamelhakham tested it in their CI. I'd still advise to try it in staging CI rather than live.
Addresses https://github.com/ARMmbed/mbed-cli/issues/660

CC @teetak01 @yogpan01

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -42,7 +42,7 @@ import shutil
 import stat
 import errno
 import ctypes
-from itertools import chain, repeat
+import time
 import zipfile
 import argparse
 
@@ -1223,7 +1223,9 @@ class Repo(object):
                         os.makedirs(os.path.split(path)[0])
 
                     info("Carbon copy from \"%s\" to \"%s\"" % (cache, path))
+                    self.cache_lock(url)
                     shutil.copytree(cache, path)
+                    self.cache_unlock(url)
 
                     with cd(path):
                         scm.seturl(formaturl(url, protocol))
@@ -1253,7 +1255,9 @@ class Repo(object):
             self.url = url
             self.path = os.path.abspath(path)
             self.ignores()
+            self.cache_lock(url)
             self.set_cache(url)
+            self.cache_unlock(url)
             return True
 
         if offline:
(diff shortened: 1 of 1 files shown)


mbed/mbed.py
```diff
@@ -1336,6 +1342,103 @@ def set_cache(self, url):
...
+
+        for i in range(timeout): 
+            if i:
+                time.sleep(1)
+
+            if os.path.exists(lock_dir):
+                try:
+                    if os.path.isfile(lock_file):
+                        with open(lock_file, 'r', 0) as f:
+                            pid = f.read(8)
```
SeppoTakalo: Concurrency issue here.
Other proces might have just opened the file, but not yet written to it.
On line 1385 it does:
```
                    with open(lock_file, 'wb', 0) as f:
                        info("Writing cache lock file %s for pid %s" % (lock_file, os.getpid()))
                        f.write(str(os.getpid()))
                        f.flush()
                        os.fsync(f)
```
So context switch might happen after `open()` but just before writing to a file.

This would lead to two processes thinking that they own the lock.

screamerbg:
Will fix the Py3 issues. Please ignore for now.

mbed/mbed.py
```diff
@@ -1223,7 +1225,9 @@ def clone(self, url, path, rev=None, depth=None, protocol=None, offline=False, *
                         os.makedirs(os.path.split(path)[0])
 
                     info("Carbon copy from \"%s\" to \"%s\"" % (cache, path))
+                    self.cache_lock(url)
                     shutil.copytree(cache, path)
+                    self.cache_unlock(url)
```
theotherjimmy: Could this be safer as a context manager? That way we could never forget to unlock the cache.

mbed/mbed.py
```diff
@@ -1253,7 +1257,9 @@ def clone(self, url, path, rev=None, depth=None, protocol=None, offline=False, *
             self.url = url
             self.path = os.path.abspath(path)
             self.ignores()
+            self.cache_lock(url)
             self.set_cache(url)
+            self.cache_unlock(url)
```
theotherjimmy: We could also use the context manager here.

teetak01:
cc: @adamelhakham

PR #753, opened 2018-09-12
Release v1.8.1

theotherjimmy:
**Fixed:**
* Concurrent Mbed CLI deploys that access the same cached repo no longer cause cache corruption.
* Python dependencies using a github url and release instead of a package name and version are checked correctly and no longer cause sperious pip installs.
* `mbed test --unittest` no longer requires a target and toolchain, which it simply throws away.

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -51,7 +51,7 @@ from contextlib import contextmanager
 
 
 # Application version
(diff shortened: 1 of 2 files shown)


theotherjimmy:
@screamerbg Is there anything else you wanted in 1.8.1?

PR #758, opened 2018-09-14
Add warning if there is no TEST_APPS folder

OPpuolitaival:
Print warning if current folder does not contain TEST_APPS folder. Fix for https://github.com/ARMmbed/mbed-os/issues/8120


theotherjimmy:
This is definitely odd, but I don't see any issues with it.

PR #765, opened 2018-10-01
Return None from MbedTerminal Constructor

### LINKED ISSUE #688: [mbed] ERROR: Unknown Error: __init__() should return None, not 'bool'
New error is shown when attempting to program the platform with a terminal application already connected: "[mbed] ERROR: Unknown Error: __init__() should return None, not 'bool'"

Mbed CLI 1.7.2

Steps to reproduce:
```
mbed import mbed-os-example-blinky
cd mbed-os-example-blinky
mbed compile -t GCC_ARM -m K64F -f

...
Total Static RAM memory (data + bss): 12352 bytes
Total Flash memory (text + data): 51513 bytes

Image: ./BUILD/K64F/GCC_ARM/mbed-os-example-blinky.bin
[mbed] Detected "K64F" connected to "/Volumes/DAPLINK" and using com port "/dev/tty.usbmodem1412"
[mbed] ERROR: Unknown Error: __init__() should return None, not 'bool'
```

theotherjimmy:
Since we examine `self.serial` to determine the success of the 
constructor, I made all `return False` statements within the constructor
set `self.serial = None`. That way the current code should catch the
failures without any additional changes

Fixes #688

### DIFF
--- a/mbed/mbed_terminal.py
+++ b/mbed/mbed_terminal.py
@@ -35,7 +35,8 @@ class MbedTerminal(object):
         try:
             from serial import Serial, SerialException
         except (IOError, ImportError, OSError):
-            return False
+            self.serial = None
+            return
 
         try:
             self.serial = Serial(self.port, baudrate=self.baudrate, timeout=self.timeout)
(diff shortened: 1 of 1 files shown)


theotherjimmy:
@screamerbg This makes the current behavior not produce the warning. I'm not sure if you wanted to distinguish between these failure modes. If you do, we may have to change the API of the MbedTerminal class to distinguish these errors.

PR #766, opened 2018-10-03
Use bytes or %r to handle paths and process output

### LINKED ISSUE #710: utf8 error by importing mercurial repository
When importing repository in the os.mbed.com (Mercurial), I got following problem in my PC (Japanese Windows 7).

```
PS C:\Users\toyom> mbed --version
1.7.3
PS C:\Users\toyom> mbed import https://os.mbed.com/teams/mbed-os-examples/code/mbed-os-example-blinky/
[mbed] Importing program "mbed-os-example-blinky" from "https://os.mbed.com/teams/mbed-os-examples/code/mbed-os-example-blinky" at latest revision in the current branch
[mbed] ERROR: Unknown Error: 'utf8' codec can't decode byte 0x91 in position 0: invalid start byte
---
```

There is no problem importing GitHub repositiry.

```
PS C:\Users\toyom> mbed import https://github.com/ARMmbed/mbed-os-example-blinky
[mbed] Importing program "mbed-os-example-blinky" from "https://github.com/ARMmbed/mbed-os-example-blinky" at latest revision in the current branch
[mbed] Adding library "mbed-os" from "https://github.com/ARMmbed/mbed-os" at rev #03196b244ee0
PS C:\Users\toyom>
```

Also, mbed-cli v1.6.0 did not have this problem.

```
PS C:\Users\toyom> mbed --version
1.6.0
PS C:\Users\toyom> mbed import https://os.mbed.com/teams/mbed-os-examples/code/mbed-os-example-blinky
[mbed] Importing program "mbed-os-example-blinky" from "https://os.mbed.com/teams/mbed-os-examples/code/mbed-os-example-blinky" at latest revision in the current branch
[mbed] Adding library "mbed-os" from "https://github.com/ARMmbed/mbed-os" at rev #03196b244ee0
PS C:\Users\toyom>
```

theotherjimmy:
Paths are not ASCII, paths are not Unicode. Paths are bytes.
This PR handles paths like they are bytes, well at least the 
current working directory. Handling paths and process output as
if they are unicode will introduce errors when paths are not 
valid unicode. I tested this by creating a directory named
    
    �abc\uXXXX

Which is not valid unicode, and tricks most python string 
decoders into throwing an exception.

Fixes #710, maybe

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -156,7 +156,10 @@ _cwd = os.getcwd()
 
 # Logging and output
 def log(msg, is_error=False):
-    sys.stderr.write(msg) if is_error else sys.stdout.write(msg)
+    if sys.version_info[0] > 2 and isinstance(msg, bytes):
+        sys.stderr.buffer.write(msg) if is_error else sys.stdout.buffer.write(msg)
+    else:
+        sys.stderr.write(msg) if is_error else sys.stdout.write(msg)
 
 def message(msg):
     if very_verbose:
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -274,7 +274,7 @@ def pquery(command, output_callback=None, stdin=None, **kwargs):
     stdout, _ = proc.communicate(stdin)
 
     if very_verbose:
-        log(stdout.strip() + "\n")
+        log(stdout.strip() + b'\n')
 
     if proc.returncode != 0:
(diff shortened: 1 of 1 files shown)


theotherjimmy:
I should get it working with python3 as well.

theotherjimmy:
This is, well very involved. Paths and other bytes being treated interchangeably with strings is everywhere in this code base.

theotherjimmy:
Closing in favor of #767

PR #775, opened 2018-10-31
Fix: Support for package types in requirements.txt

screamerbg:
Add support for package types in requirements.txt. Also fix python error when no requirements.txt file is present.

See https://github.com/ARMmbed/mbed-os/pull/8446

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1601,7 +1601,7 @@ class Program(object):
     def requirements_contains(self, library_name):
         req_path = self.get_requirements()
         if not req_path:
-            return None
+            return False
 
         req_file = 'requirements.txt'
         with open(os.path.join(req_path, req_file), 'r') as f:
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1599,13 +1599,19 @@ class Program(object):
         return None
 
     def requirements_contains(self, library_name):
-        req_path = self.get_requirements() or self.path
+        req_path = self.get_requirements()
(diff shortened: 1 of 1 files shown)


mbed/mbed.py
```diff
@@ -1615,17 +1621,17 @@ def check_requirements(self, show_warning=False):
...
                     pkg = re.sub(r'-', '_', re.sub(r'^([^<>=@]+).*$', r'\1', line).lower())
                     pkg = re.sub(r'^(git|hg|svn|bzr)\+(.*)/([\w.-]+?)(\.(git|hg))?$', r'\3', pkg)
+                    pkg = re.sub(r'\[([\w]+)\]', '', pkg)
                     if not pkg in installed_packages:
                         missing.append(pkg)
 
                 if missing and install_requirements:
                     try:
-                        action("Auto-installing missing Python modules...")
+                        action("Auto-installing missing Python modules (%s)..." % ', '.join(missing))
```
cmonr: 👍

mbed/mbed.py
```diff
@@ -1599,13 +1599,19 @@ def _find_file_paths(self, paths, fl):
         return None
 
     def requirements_contains(self, library_name):
-        req_path = self.get_requirements() or self.path
+        req_path = self.get_requirements()
+        if not req_path:
+            return None
```
cmonr: So this function can now return None, False, or True? What are these values used for?

screamerbg: Good point!

PR #779, opened 2018-11-01
CircleCI: Tests against more Mbed OS releases

screamerbg:
This small PR adds more backwards compatibility tests against Mbed OS tools versions - 5.9.6 and 5.10.2.


cmonr:
@screamerbg Are the tests suppose to pass as-is?

screamerbg:
@cmonr See bug in mbed-os test tools described here https://github.com/ARMmbed/mbed-os/issues/8619

screamerbg:
@cmonr Any rough estimation when the python 3 fixes will be available on mbed-os master?

PR #789, opened 2018-11-14
add memory trace parser to mbed client

thinkberg:
Here is the memory trace parser embedded into the sterm functionality. It was pretty easy using the transform functionality of MiniTerm. However, it also means that the sterm basically enters a line mode. So if you expect a prompt on a single line, it may not show up when using trace.

Run the terminal with `--trace <resetstring>`: `mbed sterm  --trace "ubirch" -r`

The `<resetstring>` resets the counters, so you can press the reset button and it will clear the counter. Apart from that I think the screenshot is pretty self explanatory.

The output is as follows:
screenshot 2018-11-14 at 15 26 44

Enjoy,

Leo.

### DIFF
--- a/mbed/mbed_debug.py
+++ b/mbed/mbed_debug.py
@@ -1,0 +1,132 @@
+# mbed debug trace transformer, handling memory trace information
+# Author: Matthias L. Jugel (@thinkberg)
+#
+# Copyright (c) 2018 ubirch GmbH, All Rights Reserved
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+#
+# You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+# either express or implied.
+
+import re
+import sys
+
+from serial.tools.miniterm import Transform
+
+def debug(s):
+    sys.stderr.write("D "+repr(s)+"\n")
+
+class MbedTransform(Transform):
+    mem = {}
+    allocated = 0
+    reset = None
+    buffer = ""
+
+    def __init__(self, reset_str=None):
+        self.reset = reset_str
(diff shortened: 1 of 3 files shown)


theotherjimmy (approved):
Looks good to me. @screamerbg What do you think of this sterm plugin?

screamerbg:
Looks good indeed! Thanks for this @thinkberg. I'd like to involve few guys from the various teams (including OOB) to try this functionality.

@sg-, @MarceloSalazar, @maclobdell, @toyowata, @bentcooke could you guys try this cool addition to the sterm functionality?

thinkberg:
Of course, go check it. The python is not optimal yet, but I wanted to get it done in between all the other projects. There is some room for code improvements and simplification.

screamerbg:
@thinkberg Could you please elaborate about the line mode?

thinkberg:
I have to collect the output lines to do the matching. This leads to a situation where I am collecting serial output from the board until I see a CR or NL. 
If you do something in your mbed code like: `printf("input something: "); and then wait for some input of the user from the serial line, the plugin will show all lines before `input something`, but only after it sees a newline in the board output print it to the user.

thinkberg:
Or put in other words, the plugin outputs whole lines, not characters on the terminal.

mbed/mbed_debug.py
```diff
@@ -0,0 +1,132 @@
...
+        self.r_realloc = re.compile("([^#]*)#r:0x([0-9a-f]+);0x([0-9a-f]+)-0x([0-9a-f]+);(\\d+)")
+        self.r_free = re.compile("([^#]*)#f:0x([0-9a-f]+);0x([0-9a-f]+)-0x([0-9a-f]+)")
+
+    def rx(self, rx_input):
+        # collect lines
+        out = ""
+        for c in rx_input:
+            if c == '\r' or c == '\n':
+                if len(self.buffer):
+                    out += self.trace_line(self.buffer)
```
thinkberg: This is the place where it runs the data through the trace matcher. The transformer class collects data until it sees either `\r` or `\n`

screamerbg: Thanks for the reference

thinkberg:
Opinions? Recommendations? Wishes?

thinkberg:
We should come to an end here ;-)

PR #780, opened 2018-11-01
Feature: Improve awareess about current path and current program/project

screamerbg:
This small PR aims to improve the awareness of the user about the current location and current program by adding output to every `mbed` command.

Here's an example:

```
D:\Work\smcc-tests\simple-mbed-cloud-client> mbed detect
[mbed] Working path "D:\Work\smcc-tests\simple-mbed-cloud-client" (library)
[mbed] Program path "D:\Work\smcc-tests"
[mbed] No mbed targets were detected on your system.
```

and also 

```
D:\Work\smcc-tests> mbed detect
[mbed] Working path "D:\Work\smcc-tests" (program)
[mbed] No mbed targets were detected on your system.
```

PR #790, opened 2018-11-14
Feature: Don't print location for `mbed config` for easier parsing.

screamerbg:
@MarceloSalazar raised that #780 broke the parser he wrote for `mbed config -G KEY` which is used to extract configuration from Mbed CLI and use with other tools. This PR aims to fix this and restore the old behavior for `mbed config`.

CC @MarceloSalazar

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -144,6 +144,8 @@ mbed_sdk_tools_url = 'https://mbed.org/users/mbed_official/code/mbed-sdk-tools'
 # a list of public SCM service (github/butbucket) which support http, https and ssh schemas
 public_scm_services = ['bitbucket.org', 'github.com', 'gitlab.com']
 
+# commands that don't get the current work path shown
+skip_workpath_commands = ["config", "cfg", "conf"]
 
 # verbose logging
 verbose = False
@@ -3367,7 +3369,7 @@ def main():
         very_verbose = pargs.very_verbose
         verbose = very_verbose or pargs.verbose
         pathtype = Repo.pathtype(cwd_root)
-        if not sys.argv[1].lower() in ["config", "cfg", "conf"]: # `mbed config` should be parse-friendly
+        if not sys.argv[1].lower() in skip_workpath_commands:
             action('Working path \"%s\" (%s)' % (cwd_root, pathtype))
             if pathtype == "library":
                 action('Program path \"%s\"' % Program(cwd_root).path)
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -3367,9 +3367,10 @@ def main():
         very_verbose = pargs.very_verbose
         verbose = very_verbose or pargs.verbose
         pathtype = Repo.pathtype(cwd_root)
-        action('Working path \"%s\" (%s)' % (cwd_root, pathtype))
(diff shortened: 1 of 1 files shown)


mbed/mbed.py
```diff
@@ -3356,9 +3356,10 @@ def main():
         very_verbose = pargs.very_verbose
         verbose = very_verbose or pargs.verbose
         pathtype = Repo.pathtype(cwd_root)
-        action('Working path \"%s\" (%s)' % (cwd_root, pathtype))
-        if pathtype != "program":
-            action('Program path \"%s\"' % Program(cwd_root).path)
+        if not sys.argv[1].lower() in ["config"]: # `mbed config` should be parse-friendly
+            action('Working path \"%s\" (%s)' % (cwd_root, pathtype))
+            if pathtype == "library":
```
cmonr: Why did this line need to change from `==` to `!=` ?

screamerbg: Because when you use mbed CLI on a directory that cannot be associated with a project, getting "Program path" message is very confusing - it may not be program at all. Also in the case of folder, the two lines report the same folder.

cmonr: Kk. Was primarily wondering if some third pathtype was present elsewhere, but that makes sense.

mbed/mbed.py
```diff
@@ -3356,9 +3356,10 @@ def main():
         very_verbose = pargs.very_verbose
         verbose = very_verbose or pargs.verbose
         pathtype = Repo.pathtype(cwd_root)
-        action('Working path \"%s\" (%s)' % (cwd_root, pathtype))
-        if pathtype != "program":
-            action('Program path \"%s\"' % Program(cwd_root).path)
+        if not sys.argv[1].lower() in ["config"]: # `mbed config` should be parse-friendly
```
cmonr: Any reason for doing `in ["config"]` instead of `is "config"`?

cmonr: Also, what happens if a config alias is used instead?

screamerbg:
Not sure how the PR got closed...

screamerbg:
Whops, I deleted my own branch, sorry.

mbed/mbed.py
```diff
@@ -3356,9 +3356,10 @@ def main():
         very_verbose = pargs.very_verbose
         verbose = very_verbose or pargs.verbose
         pathtype = Repo.pathtype(cwd_root)
-        action('Working path \"%s\" (%s)' % (cwd_root, pathtype))
-        if pathtype != "program":
-            action('Program path \"%s\"' % Program(cwd_root).path)
+        if not sys.argv[1].lower() in ["config", "cfg", "conf"]: # `mbed config` should be parse-friendly
```
theotherjimmy: Perhaps this should be a global, so that if I add another alias for config I don't have to add it into two places. You could also remove the comment at that point.

bridadan:

mbed/mbed.py
screamerbg: Not sure what's the proposal? Modify the subcommand routine so for every command there's a list of aliases?

theotherjimmy: make `["config", "cfg", "conf"]` a global variable.

PR #810, opened 2018-12-28
scm: fix new command only use hg to clone code

kevinong0108:
add check is hg file exist or not to avoid never try to use git.
But this patch maybe only for python 3.x

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -477,6 +477,9 @@ class Hg(object):
         return True
 
     def clone(url, name=None, depth=None, protocol=None):
+        if any(os.access(os.path.join(path, name), os.X_OK) for path in os.environ["PATH"].split(os.pathsep)) == False:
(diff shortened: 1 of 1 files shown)


theotherjimmy:
@kevinong0108 How is this any different from attempting to run `hg` and failing with almost the same error?

kevinong0108:
@theotherjimmy it will break when hg clone mbed-os failed, so it never try to use git. (Windows environment)

kevinong0108:
@theotherjimmy 
If the repository is cloned, the mbed-cli will use correct tool. The problem happened on clone only for me.
And there is no rule for hg clone fail, so here use  ProcessException to notify upper try and catch.

theotherjimmy:
@kevinong0108 Mbed CLI is not the only way to clone a repo. Please update the PR with the proposed changes.

bridadan:

PR #818, opened 2019-01-08
Modify "mbed new" to create mbed_app.json with default content.

### LINKED ISSUE #648: Generate a default mbed_app.json with "mbed new" to increase config system awareness
See [handbook-#462](https://github.com/ARMmbed/Handbook/issues/462#issuecomment-374974978)

It's also one less step the user needs to complete themselves. Step being, lookup name of file to create, mbed_app.json, and the basic syntax to follow.

aashishc1988:
Modify "mbed new" to create mbed_app.json with default content.

Fixes #648

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -37,6 +37,7 @@ import sys
 import re
 import subprocess
 import os
+import json
 import platform
 import contextlib
 import shutil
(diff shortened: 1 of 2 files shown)


aashishc1988:
@screamerbg @theotherjimmy

loverdeg-ep:
#648

mbed/mbed.py
```diff
@@ -221,6 +222,13 @@ def hide_progress(max_width=80):
     if sys.stdout.isatty():
         sys.stdout.write("\r%s\r" % (' ' * max_width))
 
+def create_default_mbed_app():
+    # Default data content
+    data = {'target_overrides':{'K64F':{'platform.stdio-baud-rate': 9600}}}
```
bridadan: How would you feel about swapping the `K64F` for `*` so it would be generic for all targets?

```suggestion
    data = {'target_overrides':{'*':{'platform.stdio-baud-rate': 9600}}}
```

theotherjimmy: @bridadan There are targets that have non-9600 default baud rates because they can't support 9600, I think.

bridadan: Ah good point (pretty sure some silicon lab targets do that).

bridadan:

Issue #816, opened 2019-01-03
export -i make_gcc_arm build fails


jahess:
Exporting mbed-os-example-mesh-minimal with -i make_gcc_arm and -m KW24D results in a project that will not link.  The error from make is:

/Users/jahess/opt/gcc-arm-none-eabi-7-2018-q2-update/bin/../lib/gcc/arm-none-eabi/7.3.1/../../../../arm-none-eabi/bin/ld: mbed-os-example-mesh-minimal.elf section `.text' will not fit in region `m_text'
/Users/jahess/opt/gcc-arm-none-eabi-7-2018-q2-update/bin/../lib/gcc/arm-none-eabi/7.3.1/../../../../arm-none-eabi/bin/ld: region m_text overflowed with text and data
/Users/jahess/opt/gcc-arm-none-eabi-7-2018-q2-update/bin/../lib/gcc/arm-none-eabi/7.3.1/../../../../arm-none-eabi/bin/ld: region `m_text' overflowed by 228812 bytes
collect2: error: ld returned 1 exit status
make: *** [mbed-os-example-mesh-minimal.elf] Error 1

The mbed_app.json was modified to work with the KW24D as shown below.  This same setup works with "mbed compile."  Obviously there are a compiler/linker option getting dropped.  The include directory list for individual files is also extremely long.

For the successful mbed compile the following global options were in affect:
GNU_ARM_PATH=/Users/jahess/opt/gcc-arm-none-eabi-6-2017-q2-update/bin
TOOLCHAIN=GCC_ARM

```
diff --git a/mbed_app.json b/mbed_app.json
index 8de7119..df32d68 100644
--- a/mbed_app.json
+++ b/mbed_app.json
@@ -1,5 +1,9 @@
 {
     "config": {
+        "storage-device":{
+            "help": "Values for storage (Thread) devices are: MESH_NVM_HEAP, MESH_NVM_SD_CARD and MESH_NVM_NONE",
+            "value": "MESH_NVM_NONE"
+        },
         "enable-led-control-example": true,
         "LED": "NC",
         "BUTTON": "NC",
@@ -7,17 +11,15 @@
     },
     "target_overrides": {
         "*": {
-            "nanostack.configuration": "lowpan_router",
-            "nsapi.default-mesh-type": "LOWPAN",
-            "mbed-mesh-api.6lowpan-nd-panid-filter": "0xffff",
-            "mbed-mesh-api.6lowpan-nd-channel-page": 0,
[...]

PR #820, opened 2019-01-09
Add --profile help text to export

bridadan:
Fixes an issue brought up in #815 (and looks like #816 as well). The `--profile` help text was missing from export, so I've added it.

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1360,7 +1360,7 @@ class Repo(object):
         lock_file = os.path.join(lock_dir, 'pid')
         timeout = 300
 
-        for i in range(timeout): 
+        for i in range(timeout):
             if i:
                 time.sleep(1)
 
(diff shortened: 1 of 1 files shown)


mbed/mbed.py
```diff
@@ -3011,6 +3011,7 @@ def dev_mgmt(toolchain=None, target=None, source=False, profile=False, build=Fal
     dict(name=['-i', '--ide'], help='IDE to create project files for. Example: UVISION4, UVISION5, GCC_ARM, IAR, COIDE'),
     dict(name=['-m', '--target'], help='Export for target MCU. Example: K64F, NUCLEO_F401RE, NRF51822...'),
     dict(name='--source', action='append', help='Source directory. Default: . (current dir)'),
+    dict(name=['--profile'], action='append', help='Path of a build profile configuration file. Default: mbed-os/tools/profiles/debug.json'),
```
theotherjimmy: I would like the default to say "debug". That's the option that users are likely to use.

bridadan: By "default" do you mean add that as part of the args? So like:
```suggestion
    dict(name=['--profile'], default='debug', action='append', help='Path of a build profile configuration file. Default: mbed-os/tools/profiles/debug.json'),
```

theotherjimmy: Nope. I was thinking in the text like:

```suggestion
    dict(name=['--profile'], action='append', help='Path of a build profile configuration file. Default: debug'),
```

bridadan: Ahh I see, would you like me to update the other `--profile` help texts in the other commands?

PR #821, opened 2019-01-09
Add future dependency

bridadan:
The Python 2 tests are currently failing for all PRs because mbed-cli uses the `builtins` module that comes with the `future` package. This was not listed in the `install_requires`, so I've added it.

FYI @screamerbg @theotherjimmy


bridadan:
Oh hang on a minute, this isn't failing for this reason. Its failing because it isn't installing the mbed os python dependencies.

PR #824, opened 2019-01-14
Include tests in PyPI tarball

dotlambda:
This is useful for downstream distributions to test if the package works correctly with their build recipe. Also, it provides a way to test if the dependencies as they are packaged by the distribution are compatible.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include README.md LICENSE pypi_readme.rst
+graft test


bridadan:
Hi @dotlambda, could you explain your use case a bit more? Which distribution in particular are you interested in supporting? And how will you be running the tests? If we do merge this, I'd like to be able to make sure we don't break how you run tests as well. Thanks!

theotherjimmy:
seems fine to me too.

Issue #756, opened 2018-09-13
`mbed compile` can access the network, through pip


theotherjimmy:
Mbed CLI calls out to Mbed OS build tools delivered with Mbed OS, and is responsible for making sure that they have all of their requirements. The following subcommands could update Mbed OS, which would change the build tools requirements. Therefore the following subcommands _should_ install requirements:
 * `mbed new`
 * `mbed import`
 * `mbed add`
 * `mbed deploy`
 * `mbed update`

The following subcommands are not supposed to interact with the network to allow offline operation, and currently might access the network through pip:
 * `mbed compile`
 * `mbed test`
 * `mbed export`

These commands should not auto-install, so that offline work is not interrupted.

@screamerbg What do you think?

screamerbg:
@theotherjimmy Agreed. Thanks for highlighting this. I'll issue a PR later today to address this.

ciarmcom:
ARM Internal Ref: MBOTRIAGE-1653

theotherjimmy:
#833 Resolved this.

Issue #795, opened 2018-11-28
Auto-installing missing Python modules


janjongboom:
Every time when I compile, I get:

```
[mbed] Auto-installing missing Python modules...
```

Which takes multiple seconds. Running Mbed OS 5.10, Mbed CLI 1.8.2.

ciarmcom:
*Internal Jira reference: https://jira.arm.com/browse/MBOCUSTRIA-200*

0xc0170:
Same here

Running master as of today (noticed this even in the last patch releases), Mbed CLI 1.8.2

Why are we getting this warning?

jeromecoutant:
Hi

See #782 
I think @janjongboom issue is the same

mbed CLI 1.8.3 is correcting this, but as tested in #782, crash can happen

0xc0170:
So this is a duplicate. @theotherjimmy / @bridadan Shall we close this?

jeromecoutant:
Proposition ?
- at first build, pip install requirements.txt is called with force update option
- requirements file is copied into BUILD directory (to compare files for next builds)
- pip is not called any more until the file is not changed

=> no more long "pip list" check, no more requirement parsing in script

Thx

JanneKiiskila:
Please fix this ASAP, it's a really annoying feature and I can only guess how much useless cycles it uses in our CIs.

screamerbg:
The python modules that are required for Mbed OS tools to run are changing over time, where more and various functionalities are added, e.g. icetea, mbed_cloud_sdk, manifest tool, but also python library modules.

To satisfy these changing requirements, it's part of the user flows (not CI flows) that Mbed CLI will check whether all required modules are installed and available for the user be able to run normally the Mbed OS tools - compile, test, export, etc. Therefore this is core part of the process and we cannot install requirements ONCE.

For your CI there's already a `--no-requirements` command switch. In the next hour I'll put up a PR that offers a configuration switch to disable the check and installation of requirements on config levels for users like yourself who love to have control of every single module in their space.

JanneKiiskila:
Those changes won't come in every day or every compile. If something breaks, we should check requirements then to fix it. The default behaviour should not be we check, but rather check when asked (in my opinion).

Issue #782, opened 2018-11-06
MBED python requirements issues ?


jeromecoutant:
Hi

Since https://github.com/ARMmbed/mbed-os/pull/8542,
python modules version check is more strict.
But I think this doesn't really work ?

1. missing module check

In my environment, urllib3 is installed.
But if I print pkg here:
https://github.com/ARMmbed/mbed-cli/blob/e99326300cc548b632132ba2e08da3fdc1c74e71/mbed/mbed.py#L1626

=> I always got urllib3[secure]

2. module version check

"pip install -r requirements.txt" command is executed only if "missing" is not empty:
https://github.com/ARMmbed/mbed-cli/blob/e99326300cc548b632132ba2e08da3fdc1c74e71/mbed/mbed.py#L1628

=> this means that if each package is present, module version is not checked
and https://github.com/ARMmbed/mbed-os/pull/8542 is not useful?

ciarmcom:
*Internal Jira reference: https://jira.arm.com/browse/MBOCUSTRIA-139*

screamerbg:
Fix is available on mbed-cli master. Could you please try it out and let me know if that fixes the problem?

jeromecoutant:
Hi

1.8.3 release is not ok

### TEST

$ pip install colorama==0.3.9
$ mbed test xxx

OK

$ pip install colorama -U
Successfully installed colorama-0.4.0
$ mbed test xxx

==>  It crashes….

````
Traceback (most recent call last):
  File "C:\Python27\Scripts\mbedgt-script.py", line 6, in <module>
    from pkg_resources import load_entry_point
  File "c:\python27\lib\site-packages\pkg_resources\__init__.py", line 3123, in <module>
    @_call_aside
  File "c:\python27\lib\site-packages\pkg_resources\__init__.py", line 3107, in _call_aside
    f(*args, **kwargs)
  File "c:\python27\lib\site-packages\pkg_resources\__init__.py", line 3136, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "c:\python27\lib\site-packages\pkg_resources\__init__.py", line 580, in _build_master
    return cls._build_from_requirements(__requires__)
  File "c:\python27\lib\site-packages\pkg_resources\__init__.py", line 593, in _build_from_requirements
    dists = ws.resolve(reqs, Environment())
  File "c:\python27\lib\site-packages\pkg_resources\__init__.py", line 781, in resolve
[...]

PR #833, opened 2019-02-01
Feature: Requirements config option and handling

screamerbg:
This PR tunes how Mbed OS python requirements are being installed:
* Add support for (global & local) config flag 'no_requirements', which disables the built in check for installed python modules.
* Mbed CLI should attempt to auto-install requirements when:
   * Codebased is being imported (`mbed import`, `mbed add`, `mbed deploy`, `mbed new`)
   * Codebased is being updated/checked out (`mbed update`)
   * Certificates are being installed or changed (`mbed dm`)
* Mbed CLI should only check and NOT try to auto-install requirements when:
   * Code is being compiled, tested or exported
   * Any Mbed CLI operation that doesn't imply network connectivity

As thoroughly described here https://github.com/ARMmbed/mbed-cli/issues/756.

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -53,7 +53,7 @@ from contextlib import contextmanager
 
 
 # Application version
-ver = '1.8.4'
+ver = '1.8.3'
 
 # Default paths to Mercurial and Git
 hg_cmd = 'hg'
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1618,7 +1618,7 @@ class Program(object):
         with open(os.path.join(req_path, req_file), 'r') as f:
             return library_name in f.read()
 
-    def check_requirements(self, show_warning_only=False):
+    def check_requirements(self, require_install=False):
         skip_requirements = self.get_cfg("NO_REQUIREMENTS", False)
         if skip_requirements:
             action("Skipping installed requirements check due to configuration flag.")
@@ -1641,7 +1641,7 @@ class Program(object):
                     if not pkg in installed_packages:
                         missing.append(pkg)
 
-                if missing and install_requirements and not show_warning_only:
+                if missing and install_requirements and require_install:
                     try:
                         action("Auto-installing missing Python modules (%s)..." % ', '.join(missing))
                         pquery([python_cmd, '-m', 'pip', 'install', '-q', '-r', os.path.join(req_path, req_file)])
@@ -1661,10 +1661,10 @@ class Program(object):
             elif os.name == 'posix':
                 msg += "\nOn Posix systems (Linux, etc) you might have to switch to superuser account or use \"sudo\""
 
-            if show_warning_only:
-                warning(msg)
-            else:
+            if require_install:
                 error(msg, 1)
+            else:
+                warning(msg)
 
         return True
 
@@ -1686,7 +1686,7 @@ class Program(object):
             shutil.copy(os.path.join(mbed_tools_path, 'default_settings.py'), os.path.join(self.path, 'mbed_settings.py'))
 
         if check_reqs:
-            self.check_requirements()
+            self.check_requirements(True)
 
     def add_tools(self, path):
         if not os.path.exists(path):
@@ -2637,7 +2637,7 @@ def compile_(toolchain=None, target=None, profile=False, compile_library=False, 
     args = remainder
     # Find the root of the program
     program = Program(getcwd(), True)
-    program.check_requirements(True)
+    program.check_requirements()
     # Remember the original path. this is needed for compiling only the libraries and tests for the current folder.
     orig_path = getcwd()
     orig_target = target
@@ -2799,7 +2799,7 @@ def test_(toolchain=None, target=None, compile_list=False, run_list=False,
     args = remainder
     # Find the root of the program
     program = Program(getcwd(), True)
-    program.check_requirements(True)
+    program.check_requirements()
     # Check if current Mbed OS support icetea
     icetea_supported = program.requirements_contains('icetea')
 
@@ -2982,7 +2982,7 @@ def test_(toolchain=None, target=None, compile_list=False, run_list=False,
 def dev_mgmt(toolchain=None, target=None, source=False, profile=False, build=False):
     orig_path = getcwd()
     program = Program(getcwd(), True)
-    program.check_requirements()
+    program.check_requirements(True)
     with cd(program.path):
         tools_dir = program.get_tools()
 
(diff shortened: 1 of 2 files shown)


screamerbg:
Related to #795, #782 and #756

mbed/mbed.py
```diff
@@ -1618,7 +1618,12 @@ def requirements_contains(self, library_name):
         with open(os.path.join(req_path, req_file), 'r') as f:
             return library_name in f.read()
 
-    def check_requirements(self, show_warning=False):
+    def check_requirements(self, show_warning_only=False):
+        skip_requirements = self.get_cfg("NO_REQUIREMENTS", False)
+        if skip_requirements:
+            action("Skipping installed requirements check due to configuration flag.")
```
theotherjimmy: Could we make this a verbose only print?

setup.py
```diff
@@ -18,7 +18,7 @@
 
 setup(
     name="mbed-cli",
-    version="1.8.3",
+    version="1.8.4",
```
theotherjimmy: It this supposed to be here?

screamerbg: Yes. Read the description. This was supposed to get in first for a quick 1.8.4 release to hotfix the issue. Due to the new feature now merged in, this will have to become 1.9.0

mbed/mbed.py
screamerbg: This is a non standard behavior and it's very easy to forget it. An extra line is not an overkill and it passes the responsibility to the user to handle requirements. Without the notice, there's no indication that one of the fundamental mbed CLI functions is not being performed because of user config.

bridadan: I think having it under a verbose print makes sense, we already tell the user to run with `-v` if the subprocess exits with a non zero code: https://github.com/ARMmbed/mbed-cli/blob/master/mbed/mbed.py#L3388

mbed/mbed.py
```diff
@@ -53,7 +53,7 @@
 
 
 # Application version
-ver = '1.8.3'
+ver = '1.8.4'
```
bridadan: We should put this in a separate PR when we're ready to make a release.

mbed/mbed.py
```diff
@@ -2977,7 +2982,7 @@ def test_(toolchain=None, target=None, compile_list=False, run_list=False,
 def dev_mgmt(toolchain=None, target=None, source=False, profile=False, build=False):
     orig_path = getcwd()
     program = Program(getcwd(), True)
-    program.check_requirements(True)
+    program.check_requirements()
```
bridadan: This is a behavior change, this will now cause the CLI to halt and exit with a non-zero return code if the requirements aren't present before running the `mbed dm` command.

I'm not necessarily against this, it just wasn't documented in the PR description or the commit message so I missed this at first.

mbed/mbed.py
```diff
@@ -1681,7 +1686,7 @@ def post_action(self, check_reqs=True):
             shutil.copy(os.path.join(mbed_tools_path, 'default_settings.py'), os.path.join(self.path, 'mbed_settings.py'))
 
         if check_reqs:
-            self.check_requirements(True)
+            self.check_requirements()
```
bridadan: This is a behavior change, this will now cause the CLI to halt and exit with a non-zero return code if the requirements aren't present after cloning (and auto-installing) mbed-os.

I'm not necessarily against this, it just wasn't documented in the PR description or the commit message so I missed this at first.

mbed/mbed.py
```diff
@@ -1636,7 +1641,7 @@ def check_requirements(self, show_warning=False):
                     if not pkg in installed_packages:
                         missing.append(pkg)
 
-                if missing and install_requirements:
+                if missing and install_requirements and not show_warning_only:
```
bridadan: Oh hang on, so now the `show_warning_only` is being used to select if things should be auto installed? That doesn't sound right.

mbed/mbed.py
screamerbg: And I already explained that not reminding that we're skipping all requirements check (which is NOT RECOMMENDED) should be clearly visible. Read my comment.

mbed/mbed.py
screamerbg: show_warning_only = do not auto-install

theotherjimmy: @screamerbg Perhaps it could use a different name then?

screamerbg: @theotherjimmy Proposal?

theotherjimmy: `require_install`?and invert the logic so that True raises an error and False warns.

mbed/mbed.py
screamerbg: To clarify the behavior:
* Mbed CLI should attempt to auto-install requirements when:
   * Codebased is being imported (`mbed import`, `mbed add`, `mbed deploy`, `mbed new`)
   * Codebased is being updated/checked out (`mbed update`)
   * Certificates are being installed or changed (`mbed dm`)
* Mbed CLI should only check and NOT try to auto-install requirements when:
   * Code is being compiled, tested or exported
   * Any Mbed CLI operation that doesn't imply network connectivity

As thoroughly described here https://github.com/ARMmbed/mbed-cli/issues/756. Also as mentioned in my first comment to this PR.

> Related to #795, #782 and #756

mbed/mbed.py
screamerbg: To clarify the behavior:
* Mbed CLI should attempt to auto-install requirements when:
   * Codebased is being imported (`mbed import`, `mbed add`, `mbed deploy`, `mbed new`)
   * Codebased is being updated/checked out (`mbed update`)
   * Certificates are being installed or changed (`mbed dm`)
* Mbed CLI should only check and NOT try to auto-install requirements when:
   * Code is being compiled, tested or exported
   * Any Mbed CLI operation that doesn't imply network connectivity

As thoroughly described here https://github.com/ARMmbed/mbed-cli/issues/756. Also as mentioned in my first comment to this PR.

> Related to #795, #782 and #756

screamerbg:
To clarify the behavior:
* Mbed CLI should attempt to auto-install requirements when:
   * Codebased is being imported (`mbed import`, `mbed add`, `mbed deploy`, `mbed new`)
   * Codebased is being updated/checked out (`mbed update`)
   * Certificates are being installed or changed (`mbed dm`)
* Mbed CLI should only check and NOT try to auto-install requirements when:
   * Code is being compiled, tested or exported
   * Any Mbed CLI operation that doesn't imply network connectivity

As thoroughly described here https://github.com/ARMmbed/mbed-cli/issues/756. Also as mentioned in my first comment to this PR.

> Related to #795, #782 and #756

In addition to this, the behavior above can be overridden using the `NO_REQUIREMENTS` config flag, which is a **severe** deviation from the standard and established behavior of Mbed CLI (NOT RECOMMENDED) and it should be made clearly visible at all times.

bridadan:
@screamerbg Thanks for the clarifications, I hadn't seen that issue before, that helps give the context.

I still would prefer to have the two flags separate in `check_requirements`. If you like I could add that in a follow up PR?

screamerbg:
Hey man, no worries! I could have made that clearer, but I had to rush to meetings 😄

Issue #831, opened 2019-01-31
Random deploy problem


JanneKiiskila:
Seems there is some weird things happening with mbed deploy, it's now failing. With mbed deploy --verbose it actually tells this:

```
jankii01@ubuntu:~/mbed/mbed-cloud-client-example-internal$ mbed deploy --verbose
[mbed] Working path "/home/jankii01/mbed/mbed-cloud-client-example-internal" (program)
[mbed] Updating library "pal-platform" to rev #bc8a1d84d7bf (tag: 2.2.0-RC1)
[mbed] Fetching revisions from remote repository to "pal-platform"
[mbed] Exec "git fetch --all --tags" in "/home/jankii01/mbed/mbed-cloud-client-example-internal/pal-platform"
Fetching origin
[mbed] Checkout "bc8a1d84d7bff7cb6a523fa03e353e23b4c2ae00" in pal-platform
[mbed] Revision "bc8a1d84d7bff7cb6a523fa03e353e23b4c2ae00" matches a branch "origin/HEAD" reference. Re-associating with branch
[mbed] Exec "git checkout origin/HEAD -q" in "/home/jankii01/mbed/mbed-cloud-client-example-internal/pal-platform"
[mbed] Updating library "mbed-os" to rev #a8f0c33eaa2c (tag: mbed-os-5.11.2)
[mbed] Fetching revisions from remote repository to "mbed-os"
[mbed] Exec "git fetch --all --tags" in "/home/jankii01/mbed/mbed-cloud-client-example-internal/mbed-os"
Fetching origin
From https://github.com/ARMmbed/mbed-os
 ! [rejected]              latest     -> latest  (would clobber existing tag)
error: Could not fetch origin
```

That actually explains why it goes wrong, but this info should be shown also w/o --verbose (rejected would clobber) as that explains what this is about.

I think the root cause is now the usage of a tag to indicate something that moves, which should not be done @adbridge @0xc0170  - tags are forever, you do not - **ever** - change them. Plus you do not make a tag and branch with the same name. That's a bug in git, if you ask me - it should decline to add either a tag or branch with the same name, if one already exists.

You can have a branch called "latest" and that can keep moving, but a tag cannot move. Ref: git documentation on [tagging](https://git-scm.com/docs/git-tag#_discussion).

ciarmcom:
*Internal Jira reference: https://jira.arm.com/browse/MBOCUSTRIA-822*

screamerbg:
[...]

PR #834, opened 2019-02-01
Fix: Add --force flag to `git fetch` to allow tags to be fetched in git 2.20

screamerbg:
Add --force flag to `git fetch` to allow tags to be fetched in git 2.20+ (related to the moving 'latest' branch in Mbed OS).

Also limit the remote repos used during fetch and push to the default one (remove `--all` flag from `git push` and `git fetch`).

See https://github.com/ARMmbed/mbed-cli/issues/831#issuecomment-459670021

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -738,7 +738,7 @@ class Git(object):
 
     def publish(all_refs=None):
         if all_refs:
-            popen([git_cmd, 'push'] + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
+            popen([git_cmd, 'push', '--all'] + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
         else:
             remote = Git.getremote()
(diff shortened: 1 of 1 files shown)


mbed/mbed.py
```diff
@@ -753,7 +753,7 @@ def publish(all_refs=None):
 
     def fetch():
         info("Fetching revisions from remote repository to \"%s\"" % os.path.basename(getcwd()))
-        popen([git_cmd, 'fetch', '--all', '--tags'] + (['-v'] if very_verbose else ([] if verbose else ['-q'])))
```
bridadan: Could you comment on why `--all` is being dropped for both `git push` and `git fetch`? Is it fixing a bug? Is it an optimization?

bridadan:
Since fixing this is a priority, I've added the `--all` flags back for now. They can always be removed in a follow up PR.

PR #838, opened 2019-02-01
Error out right away if api key is not defined in config

aashishc1988:
Addresses https://github.com/ARMmbed/mbed-os/issues/8144

Verified that if api key is not defined, we error out 

`
C:\testCheckMbedOS\cloud_example\example_without_config\mbed-cloud-client-example>mbed device-management init -d arm.com --model-name example-app -v -q
[mbed] Working path "C:\testCheckMbedOS\cloud_example\example_without_config\mbed-cloud-client-example" (program)
[mbed] ERROR: CLOUD_SDK_API_KEY is not defined. Please see documentation
`

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -2981,9 +2981,13 @@ def dev_mgmt(toolchain=None, target=None, source=False, profile=False, build=Fal
     orig_path = getcwd()
     program = Program(getcwd(), True)
     program.check_requirements(True)
+    env = program.get_env()
     with cd(program.path):
         tools_dir = program.get_tools()
 
+    if "MBED_CLOUD_SDK_API_KEY" not in env:
+        error("CLOUD_SDK_API_KEY is not defined. Please see documentation")
+    
     script = os.path.join(tools_dir, 'device_management.py')
     if not os.path.exists(script):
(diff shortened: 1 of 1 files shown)


aashishc1988:
@bridadan @theotherjimmy

teetak01:
Why do we need to make the api-key mandatory parameter on this step? If I do not care to download the developer-certificate, this should not be required in the init.

bridadan:
@teetak01, the developer certificate is required to connect a device to Pelion. The `mbed dm init` step is supposed to give you everything you need to compile a working binary.

If you do not want to download the developer certificate, what else are you using the `mbed dm init` command for? Do you only need to create the update certificate? Are you providing your own developer certificate ahead of time? Any more information about your use case would be helpful, thanks!

teetak01:
The main idea on my side was to use mbed dm init just for preparation of the update specific components and just use my existing developer certificate. The dm init is a bit strange as it picks some certificate from the Cloud account (presuming you have several) and I think there is no control which one of those it takes.

bridadan:
I believe I've captured the intent of this PR (make the error not as bad) in this PR: https://github.com/ARMmbed/mbed-cli/pull/838.

Because @aashishc1988 is currently unavailable, I'm going to close this change in favor of the one I mentioned above.

PR #846, opened 2019-02-13
Fix mbed compile --flash

vmedcy:
Unknown error is thrown while executing mbed compile --flash:

[mbed] ERROR: Unknown Error: local variable 'baudrate' referenced before assignment

Add the required argument to compile_ method.
This also fixes another error while mbed compile --flash --baudrate 9600:

[mbed] ERROR: Unknown Error: compile_() got an unexpected keyword argument 'baudrate'

This is a regression introduced by https://github.com/ARMmbed/mbed-cli/pull/837


vmedcy:
@theotherjimmy, @screamerbg: please consider this for 1.9.1: the regression in 1.9.0 was reported by various development teams in Cypress.

screamerbg:
Thanks for reporting this. I'll address it shortly.

theotherjimmy:
Thanks. 1.9.1 is out with this patch.

PR #847, opened 2019-02-13
Release 1.9.1

screamerbg:
Based on master + #846 after it lands on master


theotherjimmy:
Good call.

theotherjimmy:
Released everywhere.

PR #857, opened 2019-02-22
Fix downloading of Mbed 2 when zip is too big

bridadan:
The Mbed 2 zip was being downloaded into memory. With the latest releases, my Windows machine failed to download it because it ran out of memory! This downloads the zip in 1 MB chunks then writes them to disk to avoid running out of memory.

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -389,10 +389,13 @@ class Bld(object):
         try:
             if not os.path.exists(rev_file):
                 action("Downloading library build \"%s\" (might take a while)" % rev)
-                outfd = open(rev_file, 'wb')
                 inurl = urlopen(url)
-                outfd.write(inurl.read())
-                outfd.close()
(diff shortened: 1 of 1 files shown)


mbed/mbed.py
```diff
@@ -389,10 +389,13 @@ def fetch_rev(url, rev):
...
                 inurl = urlopen(url)
-                outfd.write(inurl.read())
-                outfd.close()
+                with open(rev_file, 'wb') as outfd:
+                    data = None
+                    while data != '':
+                        # Download and write the data in 1 MB chunks
+                        data = inurl.read(1024 * 1024)
+                        outfd.write(data)
         except:
```
theotherjimmy: Perhaps you could add at least:
```suggestion
        except Exception:
```

While you're here?

theotherjimmy:
I was not planning on creating a release for this. @bridadan @screamerbg  Do we need to release this now?

Issue #869, opened 2019-03-14
Running in Python 3 attempts to use raw_input


arskanov:
Hi,

I ran into this issue while trying out mbed-cli 1.9.1 on python3.5. Mbed-OS version 5.11.5.
Script fails in the following way:
```
> mbed publish
[mbed] WARNING: If you're using Python 3 with Mbed OS 5.8 and earlier versions, Python errors will occur when compiling, testing and exporting
---
[mbed] Working path "/home/aramanen/work/riotIisy" (program)
[mbed] Checking for local modifications...
[mbed] Uncommitted changes in program "riotIisy" in "/home/aramanen/work/riotIisy"
[mbed] ERROR: Unknown Error: name 'raw_input' is not defined
---
[mbed] WARNING: Using Python 3 with Mbed OS 5.8 and earlier can cause errors with compiling, testing and exporting
---
```

The error is on this [this line](https://github.com/ARMmbed/mbed-cli/blob/3b43228d6f21d99a4919866c697c0f73cecee313/mbed/mbed.py#L2329), and fails since raw_input() is input() in py3.
```
    if repo.dirty():
        action("Uncommitted changes in %s \"%s\" in \"%s\"" % (repo.pathtype(repo.path), repo.name, repo.path))
        if msg:
            repo.commit(msg)
        else:
            if sys.version_info[0] == 3:
                input('Press enter to commit and publish: ')
            else:
                raw_input('Press enter to commit and publish: ')
            repo.commit()
```
Seems to make it work fine.

arskanov:
Steps to reproduce:
1. Install module via pipenv
2. clone blinky
3. Modify two files in the repository
4. Commit one of them via git, leave the other as unstaged
5. run `mbed publish`

Should mbed-cli be able to handle this situation?

Sorry about the horrible shell formatting.
```

 ~/tmp  mkdir mcl2
 ~/tmp  cd mcl2
 ~/t/mcl2  pipenv --three
Creating a virtualenv for this project…
Pipfile: /home/aramanen/tmp/mcl2/Pipfile
Using /usr/bin/python3.5m (3.5.2) to create virtualenv…
⠋ Creating virtual environment...Using base prefix '/usr'
New python executable in /home/aramanen/.local/share/virtualenvs/mcl2-BkELMTBZ/bin/python3.5m
Also creating executable in /home/aramanen/.local/share/virtualenvs/mcl2-BkELMTBZ/bin/python
Installing setuptools, pip, wheel...
[...]

PR #870, opened 2019-03-14
Use input() instead of raw_input() for py3

arskanov:
Here's an initial fix for #869.

Uses `sys.version_info` since that's used elsewhere in the script as well to differentiate python versions.


theotherjimmy:
Thanks for the fix

Issue #224, opened 2016-06-23
*.lib should not allow named branches


sg-:
We've committed to initial an mbed OS release that guarantees exact reproducibility as it always had in mbed OS 2.0. This means that mbed cli should error when it detects a branch name, tag or anything other than a hash in the *.lib file for a repository.

PR #873, opened 2019-03-20
allow tag to be used in mbed import

studavekar:
Fix  for  **IOTCORE-1104** 

Branching strategy is changing starting 5.12 for example.  `mbed import` currently doesn't allow the user to specify branches/tags. There was a check which explicitly restricting to import tags other than `latest` or `tip`  maybe there is some limitation.

This change will have minimal impact on user workflow they can continue using the same command

- Example for branch `mbed import https://github.com/ARMmbed/mbed-os-example-blinky#mbed-os-5.12 blinky-5-12 -vvv`  
- Example for tag  `mbed import https://github.com/ARMmbed/mbed-os-example-blinky#mbed-os-5.12.0-rc2 blinky-5-12-rc2 -vvv`  

@screamerbg can you please review it.

cc @theotherjimmy @bridadan @SenRamakri

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1062,7 +1062,7 @@ class Repo(object):
         if m_repo_ref:
             rev = m_repo_ref.group(3)
             if rev and not re.match(r'^([a-fA-F0-9]{6,40})$', rev):
-                error('named branches not allowed in .lib, offending lib is {} '.format(os.path.basename(lib)))
+                error('Named branches not allowed in .lib, offending lib is {} '.format(os.path.basename(lib)))
 
         if not (m_local or m_bld_ref or m_repo_ref):
             warning(
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1058,6 +1058,12 @@ class Repo(object):
         m_local = re.match(regex_local_ref, ref.strip().replace('\\', '/'))
         m_repo_ref = re.match(regex_url_ref, ref.strip().replace('\\', '/'))
         m_bld_ref = re.match(regex_build_url, ref.strip().replace('\\', '/'))
+
+        if m_repo_ref:
+            rev = m_repo_ref.group(3)
+            if rev and not re.match(r'^([a-fA-F0-9]{6,40})$', rev):
+                error('named branches not allowed in .lib, offending lib is {} '.format(os.path.basename(lib)))
+
         if not (m_local or m_bld_ref or m_repo_ref):
             warning(
                 "File \"%s\" in \"%s\" uses a non-standard .lib file extension, which is not compatible with the mbed build tools.\n" % (os.path.basename(lib), os.path.split(lib)[0]))
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1038,8 +1038,6 @@ class Repo(object):
             repo.path = os.path.abspath(path or os.path.join(getcwd(), repo.name))
             repo.url = formaturl(m_repo_ref.group(1))
             repo.rev = m_repo_ref.group(3)
-            if repo.rev and repo.rev != 'latest' and repo.rev != 'tip' and not re.match(r'^([a-fA-F0-9]{6,40})$', repo.rev):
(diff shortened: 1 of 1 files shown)


bridadan:
Also as note, this affects what `mbed add` can accept as well. Part of the Mbed CLI goals is to always create a reproducible build/source state. However, since branches (and technically tags as well) can "move", we want to be sure that when you invoke `mbed add` with a branch or tag that the actual commit is written to the `.lib` file. This way the existing "reproducible build" functionality is preserved while still enabling the behavior requested in this issue.

We tested this locally and this is the behavior that we saw, so that behavior still seems to be preserved!

studavekar:
> I did a bit more testing, and there is a change that this introduces. This allows existing `.lib` files in programs to be able to specify branches and tags instead of just commits, which is a change in behavior. I'd say this behavior change is outside the scope of this PR.

This is a generic change which allows the user to provide tags/branches across. if the user wants to point a .lib to a branch/tag he/she should have the flexibility to do so on the same lines of `mbed import`  and `mbed add`. This behavior change/ feature addition is in line with the requirement where the user should have the flexibility to provide branches/tags.

I also thought of "reproducible build" part but all the bets of off when we allow `mbed import`  and `mbed add` to specify branches.

adbridge:
> Example for tag mbed import https://github.com/ARMmbed/mbed-os-example-blinky#mbed-os-5.12.0-rc2 blinky-5-12-rc2 -vvv

What is the blinky-5-12-rc2 here? I would expect to only have to specify one tag ?

studavekar:
> What is the blinky-5-12-rc2 here? I would expect to only have to specify one tag ?

blinky-5-12-rc2 is optional positional arguments specifying the destination name where we want to import the project.

```
C:\>mbed import --help
usage: mbed import [-h] [-I] [--depth [DEPTH]] [--protocol [PROTOCOL]]
                   [--insecure] [--offline] [--no-requirements] [-v] [-vv]
                   url [path]
Imports mbed program and its dependencies from a source control based URL
(GitHub, Bitbucket, mbed.org) into the current directory or specified
path.
Use "mbed add <URL>" to add a library into an existing program.

positional arguments:
  url                   URL of the program
  path                  Destination name or path. Default: current directory.
```

studavekar:
>I did a bit more testing, and there is a change that this introduces. This allows existing .lib files in programs to be able to specify branches and tags instead of just commits, which is a change in behavior. I'd say this behavior change is outside the scope of this PR.

>The existing behavior with the error was removed in this PR, so perhaps the error can just be moved to when reading .lib files?

The behavior of what .libs can have is not documented alteast https://os.mbed.com/docs/mbed-os/v5.11/tools/working-with-mbed-cli.html or in the help text. Some background can be found in  https://github.com/ARMmbed/mbed-cli/issues/224 thanks @bridadan for providing the issue.

studavekar:
@bridadan have made suggested changes can you re-review. if a named branch is added in .lib it will error. Its inline with current behavior have made the error message more verbose than before.

```
[mbed] Working path "/home/arm/delthis2/mbed-os-example-blinky" (program)
[mbed] ERROR: named branches not allowed in .lib, offending lib is sd-driver.lib
```

mbed/mbed.py
```diff
@@ -1060,6 +1058,12 @@ def fromlib(cls, lib=None):
         m_local = re.match(regex_local_ref, ref.strip().replace('\\', '/'))
         m_repo_ref = re.match(regex_url_ref, ref.strip().replace('\\', '/'))
         m_bld_ref = re.match(regex_build_url, ref.strip().replace('\\', '/'))
+
+        if m_repo_ref:
+            rev = m_repo_ref.group(3)
+            if rev and not re.match(r'^([a-fA-F0-9]{6,40})$', rev):
+                error('named branches not allowed in .lib, offending lib is {} '.format(os.path.basename(lib)))
```
screamerbg: Could you please capitalize first letter in the error. Sorry for nitpicking - just for consistency.

studavekar: done.

PR #889, opened 2019-04-24
Parse environment markers when auto installing.

### LINKED ISSUE #888: Mbed CLI tries to auto-install even when everything is up to date (hidapi on linux)
[mbed] WARNING: Missing Python modules were not auto-installed.
       The Mbed OS tools in this program require the following Python modules: hidapi

The missing module check does not understand this line:
```
hidapi>=0.7.99,<0.8.0;platform_system!="Linux"

```

bridadan:
The following change in Mbed OS added an environment marker to
the python requirements file:
https://github.com/ARMmbed/mbed-os/pull/10096/files#diff-b4ef698db8ca845e5845c4618278f29aR24
Mbed CLI did not know how to parse this, so a warning was always
generated. This change now parses the "platform_system" marker, and
adds a mechanism to add more if needed in the future.

This fixes #888.

@teetak01 Jimmy is unavailable for the month April, can you review please?

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1002,6 +1002,26 @@ class Git(object):
             if m.group(1) == "Checking out files":
                 show_progress('Checking out', (float(m.group(3)) / float(m.group(4))) * 100)
 
+_environment_markers = {
+    "platform_system": platform.system()
+}
+
+_comparators = {
+    "!=": lambda a, b: a != b,
+    "==": lambda a, b: a == b,
+}
+
+def _eval_environment_marker(marker):
+    m = re.match(r'\s*([^!=]+)\s*([!=]+)\s*([^\s]*)', marker)
+    if m:
+        try:
+            environment_marker_value = _environment_markers[m.group(1)]
+            environment_marker_cmp = m.group(3).strip("\"'")
+            return _comparators[m.group(2)](environment_marker_value, environment_marker_cmp)
+        except KeyError as e:
+            pass
+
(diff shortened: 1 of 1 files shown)


mbed/mbed.py
```diff
@@ -1652,8 +1672,14 @@ def check_requirements(self, require_install=False):
                     pkg = re.sub(r'-', '_', re.sub(r'^([^<>=@]+).*$', r'\1', line).lower())
                     pkg = re.sub(r'^(git|hg|svn|bzr)\+(.*)/([\w.-]+?)(\.(git|hg))?$', r'\3', pkg)
                     pkg = re.sub(r'\[([\w]+)\]', '', pkg)
-                    if not pkg in installed_packages:
-                        missing.append(pkg)
+                    line_parts = line.split(";")
+
+                    for line_part in line_parts[1:]:
```
cmonr: Why was the `1:` needed?

bridadan: A python requirements file looks like this:

```
pyusb>=1.0.0,<2.0.0
hidapi>=0.7.99,<0.8.0;platform_system!="Linux"
```

`line_parts` for the first line would look like: `["pyusb>=1.0.0,<2.0.0"]`.
For the second one, it would look like `["hidapi>=0.7.99,<0.8.0", "platform_system!=\"Linux\""]`.
So the `[1:]` only selects what comes after the first `;`, if there is one, which is the environment markers that we're trying to parse.

ConradBraam:
code caters for the cases I can know of, welcome improvement.

PR #900, opened 2019-06-13
Adjust python 3 warning text and behaviour

mark-edgeworth:
Since python 3 compatibility is not yet complete I've adjusted the warning message so that it no longer mentions the mbed-os version. Also the message is only printed as a warning once at the start. Previously it appeared again in a slightly different form if mbed CLI threw any exception.

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -3426,7 +3426,7 @@ def main():
 
     # Python 3 backwards compatability warning
     if sys.version_info[0] == 3:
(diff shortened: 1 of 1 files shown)


mbed/mbed.py
```diff
@@ -3461,10 +3461,6 @@ def main():
         if very_verbose:
             traceback.print_exc(file=sys.stdout)
         error("Unknown Error: %s" % e, 255)
-    finally:
```
madchutney: This was a strange thing to do in the first case, definitely good to remove it as it was generated even if there weren't any exceptions.

PR #918, opened 2019-09-03
Moduletests command added

AnttiKauppila:
Added support for moduletests command.
Gracefully copied unittests command style and modified for moduletests purposes


AnttiKauppila:
@madchutney @mark-edgeworth Could you review this?
There is also a PR pending in Mbed OS: https://github.com/ARMmbed/mbed-os/pull/11481 if you want to test this. With this command we can easily clone UNITTESTS CI job for MODULETESTS

PR #924, opened 2019-10-14
Print tip only if verbose it not used

OPpuolitaival:
It is annoying that "-v" tip is printed in all logs even if verbose is already in use

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -3431,9 +3431,9 @@ def main():
     pargs, remainder = parser.parse_known_args()
     status = 1
 
+    very_verbose = pargs.very_verbose
+    verbose = very_verbose or pargs.verbose
     try:
-        very_verbose = pargs.very_verbose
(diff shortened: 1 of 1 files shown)


mbed/mbed.py
```diff
@@ -3431,22 +3431,23 @@ def main():
...
     except ProcessException as e:
+        tip = "" if verbose else "Tip: You could retry the last command with \"-v\" flag for verbose output\n"
+
         error(
             "\"%s\" returned error.\n"
             "Code: %d\n"
             "Path: \"%s\"\n"
-            "Command: \"%s\"\n"
-            "Tip: You could retry the last command with \"-v\" flag for verbose output\n" % (e.args[1], e.args[0], e.args[3], e.args[2]), e.args[0])
+            "Command: \"%s\"\n%s" % (e.args[1], e.args[0], e.args[3], e.args[2]), e.args[0], tip)
```
JuhPuur: tip should be initialized to some string.

PR #934, opened 2019-11-25
Set default stdio-baud-rate for all targets

MarceloSalazar:
When creating a new Mbed OS program, for some reason the default baud rate is set only for K64F.
However, we should be having a generic configuration for all platforms, hence the proposed change.

@mark-edgeworth

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -229,7 +229,7 @@ def hide_progress(max_width=80):
 def create_default_mbed_app():
     # Default data content
     if not os.path.exists(mbed_app_file_name):
-        data = {'target_overrides':{'K64F':{'platform.stdio-baud-rate': 9600}}}
+        data = {'target_overrides':{'*':{'platform.stdio-baud-rate': 9600}}}
         with open(mbed_app_file_name, "w") as mbed_app_file:
             json.dump(data, mbed_app_file, indent=4)
 


mbed/mbed.py
```diff
@@ -229,7 +229,7 @@ def hide_progress(max_width=80):
 def create_default_mbed_app():
     # Default data content
     if not os.path.exists(mbed_app_file_name):
-        data = {'target_overrides':{'K64F':{'platform.stdio-baud-rate': 9600}}}
+        data = {'target_overrides':{'*':{'platform.stdio-baud-rate': 9600}}}
```
madchutney: I don't think this is needed as the default is defined here https://github.com/ARMmbed/mbed-os/blob/7a085b472b8b0b31755732fe79d9fc54a1c20c20/platform/mbed_lib.json#L24

It would also override other target specific definitions with this app config. However, I'm not sure what the K64F is here at all, perhaps as an example?

Is there a problem you are trying to solve?

MarceloSalazar: The tools should be platform agnostic.
At the moment, if you just run `mbed new test` and see the `mbed_app.json` generated inside the project, you'll see:
```
"target_overrides": {
        "K64F": {
            "platform.stdio-baud-rate": 9600
        }
```
This is for no good reason creating a configuration for the `K64F` but many other times I just want to use other platforms.
This PR is just replacing `K64F` to `*` to be agnostic and apply the same default-configuration to all platforms.

madchutney: As it stand the K64F is a strange additional but I'd rather see it removed than making it apply to all platforms.

MarceloSalazar: Ok, shall we remove the whole line?

PR #938, opened 2019-12-04
Allow developers to use branch-names in lib-files.

teetak01:
This limitation in the tool is quite artificial and causes significant
issues for development which can span multiple repositories and branches.

Allowing this would simplify maintenance of development branches with
interdependencies.

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -1081,8 +1081,6 @@ class Repo(object):
 
         if m_repo_ref:
             rev = m_repo_ref.group(3)
-            if rev and rev != 'latest' and rev != 'tip' and not re.match(r'^([a-fA-F0-9]{6,40})$', rev):
-                error('Named branches not allowed in .lib, offending lib is {} '.format(os.path.basename(lib)))
 
         if not (m_local or m_bld_ref or m_repo_ref):
             warning(


teetak01:
@JanneKiiskila @JammuKekkonen

JanneKiiskila:
Definitely needed.
We should also all usage of tags, instead of those hashes. If tags are used correctly we have no issues (i.e. according to git rules - tags are WRITE ONCE, NEVER MODIFY). Tags are much more readable than the silly hashes. I mean, that's the whole reason why we tag the Mbed OS releases in the 1st place!

JanneKiiskila:
@madchutney  @mark-edgeworth  - to next Mbed CLI release, please.

JanneKiiskila (approved):

mbed/mbed.py
```diff
@@ -1081,8 +1081,6 @@ def fromlib(cls, lib=None):
 
         if m_repo_ref:
             rev = m_repo_ref.group(3)
-            if rev and rev != 'latest' and rev != 'tip' and not re.match(r'^([a-fA-F0-9]{6,40})$', rev):
```
madchutney: We think this causes problems up stream when differentiating between SCMs. We can't put this in without further investigation.

JanneKiiskila: You mean the mercurial? I think people have to use the right types of .lib references for the tooling, it's once again something people need to solve, rather than the tool

JanneKiiskila: We're not preventing anyone from doing sensible things, even if we remove this limitation we have now.

40Grit:
I don't want my build tool messing with either my version control nor my dependency managment.

But since mbed-cli does do these things and it will be around for a bit I think our customers and us would appreciate being able to use human readable/rememberable lib references.

Yes the commit hash is safest but we can handle ourselves.

trowbridgec:
I agree too, especially for adding the use of tags.

40Grit:
@teetak01 Would using a shallow clone add any benefit to this and for all the other issues people have when cloning Mbed OS?

JanneKiiskila:
https://github.com/ARMmbed/mbed-cli/graphs/code-frequency - yep, this is not going to happen.

image

PR #964, opened 2020-03-16
mbed compile : enable flash option with dual core targets

jeromecoutant:
mbed compile -m DISCO_H747I_CM4 -t ARM -f
is using mbedls result to detect the correct plugged target on host.

There is an issue with dual core chips,
because 2 different targets exist in the targets.json file for the same HW

Associated patch for mbed test command:
https://github.com/ARMmbed/mbed-os/pull/12630

@MarceloSalazar 
@LMESTM

### DIFF
--- a/mbed/mbed.py
+++ b/mbed/mbed.py
@@ -2785,6 +2785,10 @@ def compile_(toolchain=None, target=None, macro=False, profile=False,
                     error("The '-f/--flash' option requires that the 'mbed-greentea' python module is installed.\nYou can install mbed-greentea by running \"%s -m pip install mbed-greentea\"." % python_cmd, 1)
 
                 connected = False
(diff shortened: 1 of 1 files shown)


MarceloSalazar:
@madchutney please check this and share comments.

jeromecoutant:
@madchutney we should find something like endswith("_CM4") ?

jeromecoutant:
@madchutney 
I applied same kind of change as in https://github.com/ARMmbed/mbed-os/pull/12630

jeromecoutant:
@madchutney thx

jeromecoutant:
ping

jeromecoutant:
Please make a release version with this PR,
as it is needed with https://github.com/ARMmbed/mbed-os/pull/13014

jeromecoutant:
ping

PR #965, opened 2020-03-17
Fix UT --valgrind command

kivaisan:
Fixes the problem where `mbed test --unittests --valgrind` fails to error
`[mbed] ERROR: Unknown Error: sequence item 3: expected str instance, bool found`


kivaisan:
@AnttiKauppila please review.

petroborys (approved):

